PR #9, opened 2022-11-15
protect main branch

pjfanning:
relates to #8 

I force pushed a rebase of main so it should be ok to protect main branch again


mdedetrich (approved):
lgtm

Issue #18, opened 2023-02-13
Fix gradle plugin tests


mdedetrich:
They are currently failing

pjfanning:
These tests are passing - can we close this?

Issue #17, opened 2023-02-13
Why is there a gradle plugin for this project?


mdedetrich:
I am missing some context here, but is there a reason why there is a gradle plugin for this project specifically? @jrudolph @raboof Maybe you have some idea?

jrudolph:
It's needed if you use it from Java.

mdedetrich:
Okay I will look into it then, not exactly sure whats special about this project that we need Gradle for Java specifically

raboof:
This component relies on code generation in the projects that use it, and provides Maven, gradle and sbt plugins to achieve that. I *think* it should be possible to use the generic maven/gradle/sbt protobuf plugins and "plug in" the Akka gRPC code generation, but 'dedicated' plugins are much easier to use especially for 'simple' cases.

mdedetrich:
Thanks thats a lot more clear, in that case we should fix the tests

jrudolph:
It seems to a problem with naming versions. Missing a better tag `git describe` currently names the commits after the `forked-from-akka` tag. In some cases, the leading `f` seems to be snipped off (maybe because usually the tags are `v1.0.0`?) leading to the version starting with `orked-from...` which isn't what is expected later on... Maybe most easily we should either create a random tag that has a proper version format or make sure the `forked-from-akka` tag isn't seen in the build.

mdedetrich:
Closing this, everything is clarified and the tests have been fixed

PR #19, opened 2023-02-19
Fix tests due to not setting version correctly for Gradle

mdedetrich:
So I managed to fix the CI issues behind pekko-grpc tests not passing.

As correctly stated at https://github.com/apache/incubator-pekko-grpc/issues/17#issuecomment-1427881150 the `forked-from-akka` tag was causing issues and to solve this problem I just pushed a `v0.0.0` tag to main.

There was however another additional problem as can been seen from the PR in one case we were forgetting to set the gradle version correctly. I believe that this is an oversight and the reason why it wasn't a problem in the context of Akka is that if you don't do `-Dakka.grpc.project.version=$(cat ~/.version | sed -e s/-SNAPSHOT//)` then it just defaults to the main version derived from the git tag. In the case of Akka this would have worked because there would have been actual releases of that version which would have allowed `./gradlew` to load fine however in our case there is no such release of pekko-grpc v0.0.0.

Pinging @raboof and @jrudolph incase I might have missed something or causing some regression I am not aware of.

Resolves: https://github.com/apache/incubator-pekko-grpc/issues/18
Resolves: https://github.com/apache/incubator-pekko-grpc/issues/17

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -159,7 +159,7 @@ jobs:
           cp .jvmopts-ghactions .jvmopts
           sbt akka-grpc-codegen/publishM2 akka-grpc-scalapb-protoc-plugin/publishM2 +akka-grpc-runtime/publishM2
           cd gradle-plugin
-          ./gradlew clean publishToMavenLocal --console=plain --info --stacktrace
+          ./gradlew clean publishToMavenLocal --console=plain --info --stacktrace -Dakka.grpc.project.version=$(cat ~/.version | sed -e s/-SNAPSHOT//)
           find ~/.m2 | grep gradle
 
       - name: Test Gradle Java ${{ matrix.SCALA_VERSION }}


mdedetrich:
@pjfanning You mind looking into this as well? Once the tests are functioning for this project I want to start looking into the package renaming.

pjfanning (approved):
Lgtm

jrudolph (changes requested):
After creating the tag, testing already works without any further changes. I'm not even sure that this change has any effect...

See https://github.com/apache/incubator-pekko-grpc/actions/runs/4231314530/jobs/7349682905 which worked without any change.

mdedetrich:
> After creating the tag, testing already works without any further changes. I'm not even sure that this change has any effect...

See https://github.com/apache/incubator-pekko-grpc/actions/runs/4231314530/jobs/7349682905 which worked without any change.

I was testing this locally and it needed this change to work but I guess my branch was in some weird state.

Closing PR then, as you said its working now so its unnecessary

PR #20, opened 2023-02-21
[WIP] Testing gradle test

jrudolph:
I think #19 might not even be needed...


jrudolph:
Validated that no more changes are needed to fix Gradle tests.

PR #36, opened 2023-02-28
Use Pekko Paradox sbt plugin

mdedetrich:
Working as expected, thanks to @jrudolph for solving the circe binary compatibility issue

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -183,12 +183,13 @@ lazy val docs = Project(id = "docs", base = file("docs"))
 // Make sure code generation is run:
   .dependsOn(pluginTesterScala)
   .dependsOn(pluginTesterJava)
-  .enablePlugins(AkkaParadoxPlugin, ParadoxSitePlugin, PreprocessPlugin, PublishRsyncPlugin)
+  .enablePlugins(PekkoParadoxPlugin, ParadoxSitePlugin, PreprocessPlugin, PublishRsyncPlugin)
   .disablePlugins(MimaPlugin)
   .settings(
     name := s"$pekkoPrefix-docs",
     publish / skip := true,
     makeSite := makeSite.dependsOn(LocalRootProject / ScalaUnidoc / doc).value,
+    pekkoParadoxGithub := "https://github.com/apache/incubator-pekko-grpc",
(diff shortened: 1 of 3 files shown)


mdedetrich:
So there seems to be an issue with circe, specifically sbt plugins bringing in different versions of circe. `sbt-paradox-material-theme` brings in version 0.9.0, need to track down what other sbt plugin is bringing in a different version of circe.

mdedetrich:
So I found the conflict, its conflicting with sbt-authors which happens to use circe `0.9.x` which is ancient. I have created an upstream PR at https://github.com/jonas/paradox-material-theme/pull/35 but it seems it hasn't been updated in ages.

Pinging @jrudolph so you are aware.

mdedetrich:
> Hmm, that's unfortunate, not sure if forking makes sense because there's is so much work to be done to update it to all the latest versions... Maybe we can try making a low-effort attempt to get a new version rolling. Or, we try to work around the issue with circe which might be possible without forking (all code). Or, we abandon the theme for now and go back to the generic template until we have time later to figure out a better solution.

There already is a PR to update the the project to the latest versions (see https://github.com/jonas/paradox-material-theme/pull/34 and https://github.com/apache/incubator-pekko-sbt-paradox/issues/34), so I don't think forking is too much effort. I can look into it later on in the week if no one else picks it up.

jrudolph:
It's not about the immediate work but about what it will mean later on when we need changes. An ad-hoc single use fork should be the last resort.

mdedetrich:
From my side the bigger priority is to have the pekko projects generate docs that don't look like from the 80's. I mean at some point we may not even be using the material theme, but at least until release I don't see us needing to change to much to the material theme (if anything).

nvollmar (approved):
lgtm

jrudolph:
> From my side the bigger priority is to have the pekko projects generate docs that don't look like from the 80's.

Yes, true but let's not strive for perfection for the first release.

mdedetrich:
What I meant is that without sbt-paradox the docs don't even render correctly, so by priority I meant using pekko-sbt-paradox at all so we have something thats usable.

Its not an issue now anyways, problem has been solved.

PR #37, opened 2023-02-28
Use sbt-apache-sonatype plugin

mdedetrich:
Enable sbt-apache-sonatype plugin so pekko-grpc is ready for publishing


mdedetrich:
@jrudolph @raboof So I am at odds as to why this change is causing a regression specifically with the `Test Gradle Scala` workflow, I did a little bit of looking into it and from what I can tell one of this artifacts are being generated incorrectly

* codegen/publishM2 scalapb-protoc-plugin/publishM2 +runtime/publishM2
* cd gradle-plugin && ./gradlew clean publishToMavenLocal --console=plain --info --stacktrace

I have done some inspecting of sbt settings, and the only relevant keys that I can see which have changed is `publishTo` via the change of `sonatypeCredentialHost ` (which is intentional and the point of sbt-apache-sonatype) and `organizationName` (which shouldn't really effect anything).

The only other thing i can think of that `sbt-apache-sonatype` does that could somehow effect this is it adds extra resources to the `META-INF` and maybe this is somehow messing either with the artifact generation or with the gradle plugin code that tries to detect if it is in Java or Scala mode. I haven't looked into this theory yet.

jrudolph:
The assembly generated for `scalapb-protoc-plugin` is almost empty. It seems when running from a clean slate the assembly task fails with a `java.util.zip.ZipException: duplicate entry: META-INF/LICENSE` that is not properly propagated in sbt (bug in that version of sbt-assembly) and so a half-finished jar file is produced.

jrudolph:
Bumping sbt-assembly version seems to work locally, let's see if CI agrees.

mdedetrich:
@jrudolph Perfect, appears to be working as expected now. Feel free to review and merge.

jrudolph (approved):
lgtm

mdedetrich:
@jrudolph @pjfanning 

> java.util.zip.ZipException: duplicate entry: META-INF/LICENSE

Should we create an issue to investigate this? We need to take into account what files end up in `META-INF/LICENSE` (unless its a duplicate entry in which case its not relevant).

pjfanning:
* if the build fails, we need to investigate.
* if the build is succeeding but we suspect that 2 copies of the license end up in the zip, then we will also need to do something. Some zip libraries have options to let the library remove duplicate entries.

mdedetrich:
> if the build fails, we need to investigate.

Afaik this is not the concern

> if the build is succeeding but we suspect that 2 copies of the license end up in the zip, then we will also need to do something. Some zip libraries have options to let the library remove duplicate entries.

This is what my concern could be. If we end up having 2 different `LICENSE` files in `META-INF` (which are either now merged or one of them is dropped) then we need to look into it. If the `LICENSE` files happened to be exactly the same (and hence this bug is just an artifact of how the old sbt-assembly created jars) then we don't need to worry.

mdedetrich:
I found the problem, there was a duplicate license entry in this PR which I didn't notice. Solving it now

mdedetrich:
PR created at https://github.com/apache/incubator-pekko-grpc/pull/38

PR #38, opened 2023-03-01
Remove duplicate META-INF license entry

mdedetrich:
This PR solves the bug that was revealed in https://github.com/apache/incubator-pekko-grpc/pull/37#issuecomment-1449888452. In that PR I forgot to remove the already existing code for copying over files into META-INF so we ended up copying `LICENSE`/`NOTICE` and `DISCLAIMER` files twice.

@jrudolph Do you think there is merit in downgrading the sbt-assembly plugin incase the upgrade did any change that we are not aware of? I don't think this should be an issue, sbt people are usually very strict in not adding additional changes in future versions.


pjfanning (approved):
lgtm

jrudolph (approved):
lgtm

Let's try the new version which is much improved unless we find a blocker.

Issue #29, opened 2023-02-26
publish nightly jar to apache nexus


pjfanning:

mdedetrich:
Currently blocked by https://github.com/apache/incubator-pekko-grpc/pull/37#issuecomment-1448997163

Issue #40, opened 2023-03-02
Where do we publish the gradle plugin?


mdedetrich:
So afaik (and from some basic googling) Gradle Plugins are typically published to Gradle's official plugin repository (called plugin portal). Unlike with sbt plugins, it appears to be expected to publish plugins into their plugin portal because otherwise you have to manually add plugin repositories (see https://docs.gradle.org/current/userguide/plugins.html#sec:custom_plugin_repositories), this is even the case for releases (not just snapshots). Note that Lightbend published the plugin into Gradle's plugin portal.

So the question is from an ASF policy perspective are we allowed to publish into Gradle's plugin portal or do we have to publish it into Apache's maven repository.

pjfanning:
https://plugins.gradle.org/search?term=org.apache appears to show that no ASF projects publish directly to the official plugin repo.

I think we should just document what users need to put in their gradle files - including the 'custom' gradle repo.

mdedetrich:
If thats the case then we need to figure out how to make gradle publish gradle plugins to a non standard repo

pjfanning:
I don't know but I assume that this is not much different from sbt plugins. You publish to Nexus and users can add a 'repository' definition to their Gradle script.

pjfanning:
<img width="493" alt="image" src="https://user-images.githubusercontent.com/11783444/222420919-87e4f50d-541b-4c12-8a67-5a6874a27fec.png">

See how in the Gradle doc you linked to in the description, you can include maven repo details. With production releases of pekko-grpc, mavenCentral() will do. With pekko-grpc snapshots, users will need to link to https://repository.apache.org/content/groups/snapshots/ .

mdedetrich:
Cool, can look into it after https://github.com/apache/incubator-pekko-grpc/pull/41 is merged

mdedetrich:
I will merge this, we are forced to publish Gradle plugin in Apache's maven repo due to ASF policy

PR #41, opened 2023-03-02
Fix publish

mdedetrich:
This PR is intended to fix the currently broken publish step. Note that the PR currently disables publishing the `gradle-plugin` (see https://github.com/apache/incubator-pekko-grpc/issues/40) and will be fixed later once resolved. The PR also disables publishing docs, this needs to be implemented our way.

Resolves: https://github.com/apache/incubator-pekko-grpc/issues/29

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,62 +26,61 @@ jobs:
 
       - name: Publish
         run: |-
-          cp .jvmopts-ghactions .jvmopts
-          sbt ci-release
-          CI_RELEASE=codegen/publishSigned CI_SNAPSHOT_RELEASE=codegen/publish sbt ++2.13.8\! ci-release
+          cp .jvmopts-ci .jvmopts
+          sbt +publish
         env:
-          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
-          PGP_SECRET: ${{ secrets.PGP_SECRET }}
-          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          NEXUS_USER: ${{ secrets.NEXUS_USER }}
+          NEXUS_PW: ${{ secrets.NEXUS_PW }}
 
-  gradle-plugin:
-    name: Release gradle plugin
-    runs-on: ubuntu-latest
-    if: github.repository == 'apache/incubator-pekko-grpc'
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
-          fetch-depth: 0
+# TODO: Need to figure out where to publish gradle plugin, see https://github.com/apache/incubator-pekko-grpc/issues/40
+#  gradle-plugin:
+#    name: Release gradle plugin
+#    runs-on: ubuntu-latest
+#    if: github.repository == 'apache/incubator-pekko-grpc'
+#    steps:
+#      - name: Checkout
+#        uses: actions/checkout@v3
+#        with:
+#          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+#          fetch-depth: 0
+#
+#      - name: Setup Java 8
+#        uses: actions/setup-java@v3
+#        with:
+#          distribution: temurin
+#          java-version: 8
+#
+#      - name: Publish Plugin to Gradle Plugin Repository
+#        run: cd gradle-plugin && ./gradlew publishPlugins -Pgradle.publish.key='${{ secrets.GRADLE_PUBLISH_KEY }}' -Pgradle.publish.secret='${{ secrets.GRADLE_SECRET }}'
 
-      - name: Setup Java 8
-        uses: actions/setup-java@v3
-        with:
-          distribution: temurin
-          java-version: 8
-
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -26,62 +26,60 @@ jobs:
 
       - name: Publish
         run: |-
-          cp .jvmopts-ghactions .jvmopts
-          sbt ci-release
-          CI_RELEASE=codegen/publishSigned CI_SNAPSHOT_RELEASE=codegen/publish sbt ++2.13.8\! ci-release
+          cp .jvmopts-ci .jvmopts
+          sbt +publish
```
mdedetrich, line 30: So from what I can tell the `+publish` is all that is needed here? I did a `+publishLocal` and at least from what I could tell it works properly.

nvollmar (approved):
lgtm

.github/workflows/publish.yml
mdedetrich: @jrudolph @raboof Maybe you might have some context here

raboof: I don't have the details, but I seem to remember `+` would publish the same artifact(s) twice for some of the Scala versions, which Sonatype wouldn't accept. If we're now first publishing to an Apache repo instead of directly to Sonatype it's indeed very possible that this is no longer needed.

Also, 'publish' unlike 'ci-release' doesn't do signing etc, but that might not be a problem for the same reason.

mdedetrich: > I don't have the details, but I seem to remember `+` would publish the same artifact(s) twice for some of the Scala versions, which Sonatype wouldn't accept. If we're now first publishing to an Apache repo instead of directly to Sonatype it's indeed very possible that this is no longer needed.

Is this problem specific for this project or just in general?

> Also, 'publish' unlike 'ci-release' doesn't do signing etc, but that might not be a problem for the same reason.

Yes, this is just for snapshots which are currently not signed. Actually release is something we still need to decide upon later

raboof: > Is this problem specific for this project or just in general? Afaik `+publish` in general should just work so I guess this project has set things up differently?

I think it had to do with the fact that in this project, some submodules are released for different Scala versions than others.

> Yes, this is just for snapshots which are currently not signed. Actually release is something we still need to decide upon later

:+1:

pjfanning (approved):
Lgtm - only real way to test is to merge it

jrudolph:
The error in the failed runs points to missing nexus credentials, not sure if anything else would be needed to fix it.

jrudolph:
> The error in the failed runs points to missing nexus credentials, not sure if anything else would be needed to fix it.

Indeed, no credentials are set in this repo: https://github.com/apache/incubator-pekko-grpc/actions/runs/4311981774/jobs/7521978906#step:4:9

mdedetrich:
> The error in the failed runs points to missing nexus credentials

There is already a ticket on this

> not sure if anything else would be needed to fix it.

https://github.com/apache/incubator-pekko-grpc/pull/41/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7 There is the potential issue here. This PR is just using `+publish` which might mean we are not producing all of the relevant artifacts for specific scala versions

pjfanning:
This repo is mentioned in https://issues.apache.org/jira/browse/INFRA-24284 - so hopefully, INFRA will look at all the repos.

jrudolph:
> > not sure if anything else would be needed to fix it.
> 
> [#41 (files)](https://github.com/apache/incubator-pekko-grpc/pull/41/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7) There is the potential issue here. This PR is just using `+publish` which might mean we are not producing all of the relevant artifacts for specific scala versions

But since it worked before and there's the credential issue to fix first why change anything else right now?

mdedetrich:
> But since it worked before

I am confused, I don't think it every worked before (at least ever since the fork happened)?

jrudolph:
> I am confused, I don't think it every worked before (at least ever since the fork happened)?

I meant before the fork :) Shouldn't it work mostly the same after changing the repository to the Apache Nexus?

mdedetrich:
> I meant before the fork :) Shouldn't it work mostly the same after changing the repository to the Apache Nexus?

So what I mean is that this PR could potentially be including a regression because its changing

```
sbt ci-release
CI_RELEASE=codegen/publishSigned CI_SNAPSHOT_RELEASE=codegen/publish sbt ++2.13.8\! ci-release
```
to just
```
sbt +publish
```

And according to @raboof , the reason why current the publishing is not so trivial is because we need to publish specific sbt projects with specific scala versions (which is what the `!` syntax does, just learned this today https://discord.com/channels/632150470000902164/922600050989875282/1081143104029077545)

jrudolph:
I think I mean the same. Let's just fix the credentials first and see how much that fixes and then see what else is needed.

> which is what the `!` syntax does

Usually `!` allows to specify versions which are not listed in `crossScalaVersions`. The problem here is likely bigger than that because of the complex requirements of how some modules (e.g. sbt-plugin) can only work with some Scala versions. At least in the past, there were some tricky issues in sbt where using `++` sometimes even changes Scala versions in modules where it should not. Using `++`/`+` turned out to be somewhat brittle in those cases.

mdedetrich:
> I think I mean the same. Let's just fix the credentials first and see how much that fixes and then see what else is needed.

Cool, I will merge the PR then after we get notified that the credentials are fixed

mdedetrich:
Publishing works, 🎆

PR #50, opened 2023-03-20
Update pekko snapshot

mdedetrich:
Updates latest pekko snapshot to bring in latest ccompat changes. Although pekko-grpc doesn't need it, its best to make sure that our pekko modules have this pekko version as lowest common denominator so we don't get potential dependency resolution issues.


pjfanning (approved):
lgtm

PR #53, opened 2023-03-25
Update sbt

mdedetrich:
Update sbt version

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -21,6 +21,9 @@ addSbtPlugin("org.mdedetrich" % "sbt-apache-sonatype" % "0.1.6")
 // allow access to snapshots for pekko-sbt-paradox
 resolvers += "Apache Nexus Snapshots".at("https://repository.apache.org/content/repositories/snapshots/")
 
+// See https://github.com/akka/akka-http/pull/3995 and https://github.com/akka/akka-http/pull/3995#issuecomment-1026978593
+libraryDependencySchemes += "org.scala-lang.modules" %% "scala-xml" % "always"
+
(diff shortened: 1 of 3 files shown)


mdedetrich:
sbt 1.8.2 is more strict and because of this its complaining about some libraries, will check it out later

project/plugins.sbt
```diff
@@ -21,6 +21,9 @@ addSbtPlugin("org.mdedetrich" % "sbt-apache-sonatype" % "0.1.6")
 // allow access to snapshots for pekko-sbt-paradox
 resolvers += "Apache Nexus Snapshots".at("https://repository.apache.org/content/repositories/snapshots/")
 
+// See https://github.com/akka/akka-http/pull/3995 and https://github.com/akka/akka-http/pull/3995#issuecomment-1026978593
```
mdedetrich, line 24: This is necessary because we are getting many conflict `scala-xml` versions from various libraries. Note that this only applies to the sbt classpath for plugins and not pekko-grpc in general.

mdedetrich:
PR is now ready to review. One thing to note that is due to pekko-http-cors bringing in an incompatible version of pekko-http (see https://github.com/lomigmegard/pekko-http-cors/blob/2fad65439d2fbf9fa8ef6a63863ca04a179b7b81/build.sbt#L53) I had to make an exclusion rule which can be removed once a proper release of pekko-http and then pekko-http-cors is made.

He-Pin (approved):
LGTM，better with a //TODO in case forgetting that

mdedetrich:
> LGTM，better with a //TODO in case forgetting that

I will add TODO's in next PR

PR #61, opened 2023-04-21
standardise on scalatest 3.2.15

pjfanning:
* in pekko core, we have 'testkit' libs and those still support scalatest 3.1 for historic reasons (3.2 needed for scala 3 though).
* grpc doesn't have such a 'testkit', so seems tidier to use scalatest 3.2 - especially since pekko-http may soon support scala3 allowing pekko-grpc to potentially support scala3


raboof (approved):
Given these are all in test scope, as long as the tests succeed it seems reasonable to prefer higher versions :+1:

PR #62, opened 2023-04-28
publish grpc docs to nightlies.apache.org

pjfanning:
grpc docs are linked to from the main pekko docs - useful to have these docs published so the links work

### DIFF
--- a/.github/actions/sync-nightlies/action.yml
+++ b/.github/actions/sync-nightlies/action.yml
@@ -1,0 +1,86 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


.github/actions/sync-nightlies/action.yml
```diff
@@ -0,0 +1,86 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
mdedetrich, line 1: Since this is not source files isn't the SPDX header format enough here (i.e. `SPDX-License-Identifier: Apache-2.0`)?

pjfanning: this is a copy of a file checked into other pekko repos - it just seems easier to keep the same contents

mdedetrich (approved):
lgtm

PR #70, opened 2023-05-23
more pekko naming changes

pjfanning:
relates to #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ toolkit.
 
 ## Project Status
 
-This library is ready to be used in production, but API's and build system plugins are still expected to be improved and [may change](https://doc.akka.io/docs/akka/current/common/may-change.html).
+This library is ready to be used in production, but API's and build system plugins are still expected to be improved and [may change](https://pekko.apache.org/docs/pekko/current/common/may-change.html).
 
 The API on both sides (Client and Server) is a simple Pekko Streams-based one.
 
@@ -49,10 +49,6 @@ The project is split up in a number of subprojects:
 
 Additionally, 'plugin-tester-java' and 'plugin-tester-scala' contain an example
 project in Java and Scala respectively, with both sbt and Gradle configurations.
-
-## Compatibility & support
-
-If used with JDK 8 prior to version 1.8.0_251 you must add an [ALPN agent](https://doc.akka.io/docs/akka-http/10.1/server-side/http2.html#application-layer-protocol-negotiation-alpn-).
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ The project is split up in a number of subprojects:
 Additionally, 'plugin-tester-java' and 'plugin-tester-scala' contain an example
 project in Java and Scala respectively, with both sbt and Gradle configurations.
 
+## Compatibility & support
+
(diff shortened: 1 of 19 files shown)


pjfanning:
there are comms issues with the link validator job (the CI job keeps failing when trying to download code from Coursier)

I can double check the link validator job another day but would like to get this merged as is.

README.md
```diff
@@ -50,10 +50,6 @@ The project is split up in a number of subprojects:
 Additionally, 'plugin-tester-java' and 'plugin-tester-scala' contain an example
 project in Java and Scala respectively, with both sbt and Gradle configurations.
 
-## Compatibility & support
-
-If used with JDK 8 prior to version 1.8.0_251 you must add an [ALPN agent](https://doc.akka.io/docs/akka-http/10.1/server-side/http2.html#application-layer-protocol-negotiation-alpn-).
```
pjfanning, line 55: the link does not work with Pekko docs because that subsection was removed from Akka docs before we forked - seems tidiest just to remove this

docs/src/main/paradox/client/index.md
```diff
@@ -1,6 +1,6 @@
 # Consuming Services (Client)
 
-Get our Quickstart project (available for [Scala](https://developer.lightbend.com/start/?group=akka&project=akka-grpc-quickstart-scala) and [Java](https://developer.lightbend.com/start/?group=akka&project=akka-grpc-quickstart-java)) to get started quickly.
+See the Akka gRPC Quickstart project (available for [Scala](https://developer.lightbend.com/start/?group=akka&project=akka-grpc-quickstart-scala) and [Java](https://developer.lightbend.com/start/?group=akka&project=akka-grpc-quickstart-java)) to get started quickly.
```
pjfanning, line 4: we haven't migrated these quickstart projects yet - so seems best to admit that the Akka one is useful as a guide

mdedetrich: Fair enough, I guess we can always update this later when we migrate the docs.

README.md
mdedetrich: Rather than removing it can we instead just comment that you need to use JDK 8 1.8.0_251 or later?

mdedetrich:
> there are comms issues with the link validator job (the CI job keeps failing when trying to download code from Coursier)

I noticed these issues in general when downloading Pekko snapshot artifacts. It seems that that we are putting too much pressure on Apache Snapshots repo and in some cases the snapshots fail to download.

mdedetrich (changes requested):
Couple of remarks, otherwie lgtm.

mdedetrich:
> 

Good to know

README.md
pjfanning: I've added this back but highlight the link is an Akka link.

mdedetrich: I have a preference for JDK 8 1.8.0_251 or later because I don't want to encourage people using older JDK's and as a bonus we don't need to link to any Akka docs.

pjfanning: this same stuff is basically in the paradox docs - I'd prefer to keep them consistent

mdedetrich: Okay fine, I won't die on this hill. Will approve.

mdedetrich (approved):
lgtm

PR #71, opened 2023-05-23
remove unused plugin

pjfanning:
git.io connectivity issues are causing havoc with link validator CI builds across all our projects - the other CI checks passed

we use github workflow scripting to rsync - we don't use this sbt rsync plugin


mdedetrich (approved):
lgtm

PR #91, opened 2023-06-13
compile issue - now that we support Scala 3

pjfanning:
This implicit without a type declaration is causing major CI build problems.

Oddly the fails here are in the 2.13 builds but they do relate to this untyped implicit. https://github.com/apache/incubator-pekko-grpc/pull/90/checks


plugin-tester-scala/src/test/scala/example/myapp/helloworld/ErrorReportingSpec.scala
```diff
@@ -37,7 +37,7 @@ import scala.concurrent.duration._
 @RunWith(classOf[JUnitRunner])
 class ErrorReportingSpec extends AnyWordSpec with Matchers with ScalaFutures with BeforeAndAfterAll {
 
-  override implicit val patienceConfig = PatienceConfig(5.seconds, Span(100, org.scalatest.time.Millis))
+  override implicit val patienceConfig: PatienceConfig = PatienceConfig(5.seconds, Span(100, org.scalatest.time.Millis))
```
He-Pin, line 40: Maybe add a final too.

He-Pin (approved):
lgtm

PR #100, opened 2023-06-16
use latest gradle 6 release

pjfanning:


mdedetrich:
I was thinking about this and whether it makes sense to have different folders for the different gradle versions rather than just overriding the current one, i.e. renaming current `gradle-plugin` to `gradle-6-plugin` and adding `gradle-7-plugin`, `gradle-8-plugin` etc etc.

I think that the binaries that Gradle plugins produce are sensitive to the gradle version, so we can support multiple gradle versions for the same pekko version. We can of course drop older versions of gradle (i.e. gradle 6) in future versions of Pekko, but at least for initial release to make it easier for people to migrate we should still support gradle 6 and provide a path to later gradle versions.

pjfanning:
I don't think it is common to support many gradle versions with multiple variants of the plugin.

So far it appears that one change in gradle 7 is causing us trouble. Gradle 8 might cause even more trouble.

mdedetrich:
> I don't think it is common to support many gradle versions with multiple variants of the plugin.

Ordinarily yes but outside of Pekko I have heard horror stories of how painful it is for users to deal with gradle migrations so I think that at least for version 1.0.x of Pekko, supporting multiple gradle versions has merit. That way current Akka users who are still stuck on Gradle 6 can move to Pekko without upgrading their gradle build, and then can update to 7/8 as they please.

> So far it appears that one change in gradle 7 is causing us trouble. Gradle 8 might cause even more trouble.

Not surprising at all considering what I just mentioned.

PR #101, opened 2023-06-18
Allow producing JDK8 bytecode on newer JDK's

mdedetrich:
Still need to verify this with jardiff


raboof (commented):
looks like you're still missing the `--release` flag for the javacOptions when building with Java 9 or later.

mdedetrich:
Thanks, completely forgot about this PR. Will check it out later

project/Common.scala
```diff
@@ -79,6 +79,13 @@ object Common extends AutoPlugin {
                                      "-Wconf:msg=unused import:silent",
                                      "-Wconf:cat=feature:silent")),
     Compile / console / scalacOptions ~= (_.filterNot(consoleDisabledOptions.contains)),
+    // restrict to 'compile' scope because otherwise it is also passed to
+    // javadoc and -target is not valid there.
```
raboof, line 83:
```suggestion
    // javadoc and --release is not valid there.
```

raboof (commented):
one small suggestion, otherwise LGTM - is there a reason this PR is still in 'Draft' mode?

He-Pin:
I think we can close this, as we are moving to Java 17 on main.

raboof:
perhaps it would still be useful to produce jdk17 bytecde n newer JDK's - but that can be simpler because it doesn't need the `onlyAfterJdk8` so perhaps fine to do that in a different PR

PR #120, opened 2023-08-02
ensure old akka-grpc deprecation versions are specified to be akka-grpc

pjfanning:
* the text is consistent with existing deprecation versions (that also use 'akka-grpc')


He-Pin (approved):
Lgtm

PR #121, opened 2023-08-02
pekko http 1.0.0

pjfanning:


nvollmar (approved):
lgtm

mdedetrich (changes requested):
There are other cases of Apache snapshots repository that you can remove, i.e. https://github.com/apache/incubator-pekko-grpc/blob/719d069aa0149642fbb797c764d677ffdc9f1c88/plugin-tester-java/pom.xml#L23-L29 .

I also think that https://github.com/apache/incubator-pekko-grpc/blob/a605bcfbfd881d433fabdc12dfc05972c9f7b2b7/gradle-plugin/src/main/groovy/org/apache/pekko/grpc/gradle/PekkoGrpcPlugin.groovy#L58-L60 can be removed

PR #125, opened 2023-08-04
document how to use the gradle plugin

pjfanning:
* we don't yet have a way to publish the plugin to Gradle Plugins portal
* to unblock the pekko-grpc release, document how to build the plugin locally and how to consume it


mdedetrich (approved):
lgtm

PR #128, opened 2023-08-04
fix link in README.md

pjfanning:
the broken link to pekko-http would work if this was a Paradox markdown file - but it is not


mdedetrich (approved):
lgtm

PR #140, opened 2023-08-04
add license headers

pjfanning:
some files missing license headers


He-Pin:
I think it would be  better  to rewrite some demo from scratch.

mdedetrich (approved):
lgtm

Issue #113, opened 2023-07-18
register a gradle plugin for pekko grpc


pjfanning:
Our equivalent of https://plugins.gradle.org/plugin/com.lightbend.akka.grpc.gradle

@raboof seems to be the owner of the Akka one - maybe Arnout can help us work out what we need to do

We need to fix up the paradox docs to match whatever we end up with Gradle plugin wise.

mdedetrich:
I remember discussing this with you @pjfanning and I am under the impression that we are not allowed to publish the pekko-grpc gradle plugin into Gradles plugin repository because its against ASF policy.

Instead we need to publish the jars into ASF's nexus release repo, just like we do with sbt plugins or any other project.

pjfanning:
@mdedetrich whatever we end up with, we need to fix the paradox docs to reflect the right approach - they current say we have a Gradle plugin at the standard Gradle repo.

mdedetrich:
Agreed, this may end up taking some time/being annoying as we have to document how to publish via (i.e. doing `publishSigned` in gradle etc etc). Same deal goes with maven

mdedetrich:
Okay so regarding whether we can publish our Gradle plugin to Gradles own Plugin repository versus our ASF nexus the only thing I could find is

> Where possible use platforms officially supported by Infra.

From https://incubator.apache.org/guides/distribution.html (under Release Platforms).

mdedetrich:
Just got news from `#asfinfra` that they have a relationship with Gradle for publishing plugins, just need to make a ticket to help set it up.

mdedetrich:
INFRA ticket created https://issues.apache.org/jira/browse/INFRA-24804

pjfanning:
@mdedetrich you can take this and if you run into issues or don't have time, we can review again. I can look at the other open issues.

PR #142, opened 2023-08-07
Setup gradle plugin for publishing

mdedetrich:

Only thing that is missing is signing (see https://github.com/gradle-nexus/publish-plugin#configure-signing) but lets do that separately.

### DIFF
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -3,7 +3,7 @@ plugins {
   id 'java-gradle-plugin'
   id 'maven-publish'
   id 'com.palantir.git-version' version '0.10.1'
-  id 'com.gradle.plugin-publish' version '0.11.0'
+  id 'io.github.gradle-nexus.publish-plugin' version '1.0.0'
 }
 
 group = "org.apache.pekko"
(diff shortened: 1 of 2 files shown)


gradle-plugin/build.gradle
```diff
@@ -23,18 +23,43 @@ gradlePlugin {
...
 }
 
-pluginBundle {
-  website = 'https://pekko.apache.org/docs/pekko-grpc/current/'
-  vcsUrl = 'https://github.com/apache/incubator-pekko-grpc'
-  description = 'Building streaming gRPC servers and clients on top of Apache Pekko Streams'
-  tags = ['pekko', 'streams', 'reactive']
+publishing {
+    publications.withType(MavenPublication) {
+        pom {
```
mdedetrich, line 28: A lot of the pom details were outdated/missing. This data was retrieved by copying it from the `pom` of one of the sbt artifacts of this project

gradle-plugin/build.gradle
```diff
@@ -23,18 +23,43 @@ gradlePlugin {
...
 
-  plugins {
-    pekkoGrpcPlugin {
-      id = 'org.apache.pekko.grpc.gradle'
-      displayName = 'Apache Pekko gRPC'
+nexusPublishing {
+    repositories {
+        sonatype {
+            nexusUrl.set(uri("https://repository.apache.org/"))
+            snapshotRepositoryUrl.set(uri("https://repository.apache.org/content/repositories/snapshots/"))
```
pjfanning, line 58: so when we need a release manager (a human) to do a release, they will need to hand edit this?

pjfanning: the release URL will be something like https://repository.apache.org/service/local/staging - I need to track the correct path but it is something similar to this.

mdedetrich: > so when we need a release manager (a human) to do a release, they will need to hand edit this?

Nope, aside from maybe the version (but we have the same issue with sbt)

> the release URL will be something like https://repository.apache.org/service/local/staging - I need to track the correct path but it is something similar to this.

This should already be handled by the various `publishToSonatype` tasks and its variants as described in https://github.com/gradle-nexus/publish-plugin#configure-signing. The `https://repository.apache.org/service/local/staging` uri is nothing that should ever be set, thats an internal detail of how bundle/standard publishing is done by the plugin. If you check sbt-apache-sonatype (https://github.com/mdedetrich/sbt-apache-sonatype/blob/main/src/main/scala/org/mdedetrich/apache/sonatype/ApacheSonatypePlugin.scala#L36) you can see we only set the URI in on place.

In summary its the role of the plugins (i.e. gradle nexus publish-plugin/sbt-sonatype) to configure the staging repos, you only need to set the base URI. What may be a difference with the gradle nexus publish-plugin is that it may not support bundle publishing, but there isn't much we can do about that and its also not that critical since its just a single artifact and not many.

pjfanning: you've hardcoded a snapshot url into this publish section so I do not see how this publish can handle when we need to do a non-snapshot release - we are actually waiting to do a release on this very repo

pjfanning (changes requested):
I'm not sure at all about hardcoding snapshot urls into the gradle publishing

gradle-plugin/build.gradle
mdedetrich: This is how its documented to be done, from https://github.com/gradle-nexus/publish-plugin#publishing-to-maven-central-via-sonatype-ossrh

> Important. Users registered in Sonatype after [24 February 2021](https://central.sonatype.org/news/20210223_new-users-on-s01/) need to customize the following URLs:
>
 ```
  nexusPublishing {
    repositories {
        sonatype {  //only for users registered in Sonatype after 24 Feb 2021
            nexusUrl.set(uri("https://s01.oss.sonatype.org/service/local/"))
            snapshotRepositoryUrl.set(uri("https://s01.oss.sonatype.org/content/repositories/snapshots/"))
        }
     }
  }
  ```

pjfanning: I understand now. Thanks.

pjfanning (approved):
lgtm

PR #143, opened 2023-08-07
Fix grpc gradle sonatype publish uri

mdedetrich:
The previously entered URI is wrong, instead should have `/service/local` at the end (i.e. Apache equivalent of https://github.com/gradle-nexus/publish-plugin/blob/b116c25e9271734341be8078ab264b70e87f51cc/src/main/kotlin/io/github/gradlenexus/publishplugin/DefaultNexusRepositoryContainer.kt#L37)


pjfanning (approved):
lgtm

PR #145, opened 2023-08-07
Derive gradle version from sbt

mdedetrich:
So I can confirm that the reason behind the gradle plugin not building correctly is that it wasn't setting the version correctly (see https://github.com/apache/incubator-pekko-grpc/actions/runs/5786987516/job/15682898907). After spending the last few hours digging around to figure out how gradle publishing works I came up with an elegant solution that also simplifies whats going on.

Currently the gradle version is manually derived from executing git and parsing the results. Not only is this quite brittle (the sed regex would fail once we start releasing RC's/milestones for this same reason https://github.com/apache/incubator-pekko-http/issues/293#issuecomment-1666861760) but it also requires quite a bit of boilerplate because every single time we need to publish the gradle plugin (whether it is locally, as a snapshot or in production) we have to repeat the boilerplate of deriving the version from git tag and placing it in `~/.version` or when doing a full release setting the version in `build.gradle`.

So the best way to solve is this issue is why don't we just get the version directly from sbt? sbt is the source of truth when it comes to versions, its already doing the job deriving the version from a git tag via sbt-dynver and if the gradle plugin was hypothetically written as an sbt build it would be using the same version anyways.

If you want to verify this PR locally, you can just run `./gradlew properties -q | awk '/^version:/ {print $2}'`, `./gradlew properties` is the source of truth for what is the final calculated version.

tl;dr This change means that loading gradle will "just work" and removes all of the hacky git/regex parsing that currently exists, making the whole build a lot simpler. As a bonus it also means that the project works locally (historically I had to manually hardcode the version in build.gradle because as soon as the git pattern failed the regex the gradle build would fail to load).

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -32,7 +32,7 @@ jobs:
           NEXUS_PW: ${{ secrets.NEXUS_PW }}
 
       - name: Publish Gradle Plugin to Apache Nexus Repository
-        run: cd gradle-plugin && ./gradlew properties -q | awk '/^version:/ {print $2}' && ./gradlew publishToSonatype
+        run: cd gradle-plugin && ./gradlew publishToSonatype
         env:
           NEXUS_USER: ${{ secrets.NEXUS_USER }}
           NEXUS_PW: ${{ secrets.NEXUS_PW }}
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -2,17 +2,25 @@ plugins {
   id 'groovy'
   id 'java-gradle-plugin'
   id 'maven-publish'
-  id 'com.palantir.git-version' version '0.10.1'
   id 'io.github.gradle-nexus.publish-plugin' version '1.0.0'
 }
 
 group = "org.apache.pekko"
-// https://github.com/palantir/gradle-git-version/issues/97
 
-def tag = "git describe --tags".execute().text.substring(1).split("-g")[0].replace("\n", "")
-def finalVersion = (tag == versionDetails().lastTag.substring(1)) ? tag : tag.reverse().reverse() + "-" + versionDetails().gitHash.substring(0, 8)
-
-version = finalVersion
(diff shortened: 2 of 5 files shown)


gradle-plugin/build.gradle
```diff
@@ -2,17 +2,23 @@ plugins {
   id 'groovy'
   id 'java-gradle-plugin'
   id 'maven-publish'
-  id 'com.palantir.git-version' version '0.10.1'
```
mdedetrich, line 5: This dependency is no longer needed, it was the gradle equivalent of `sbt-dynver` but as can be seen from the referenced github issue below it didn't even work properly which is explains all of the manual git calls

.github/workflows/publish-nightly.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           NEXUS_PW: ${{ secrets.NEXUS_PW }}
 
       - name: Publish Gradle Plugin to Apache Nexus Repository
-        run: cd gradle-plugin && ./gradlew properties -q | awk '/^version:/ {print $2}' && ./gradlew publishToSonatype
```
mdedetrich, line 35: This was used to print the version, no need for this now

pjfanning (approved):
lgtm

pjfanning (changes requested):
the gradle tests are failing - eg https://github.com/apache/incubator-pekko-grpc/actions/runs/5788328185/job/15687098472?pr=145

mdedetrich:
> the gradle tests are failing - eg https://github.com/apache/incubator-pekko-grpc/actions/runs/5788328185/job/15687098472?pr=145

Yup I am looking into it

mdedetrich:
@pjfanning All is good now. The reason for the previous failing test is I used Intellij's inspections to remove unused imports in one of the files I was editing but it turns out that these imports where in fact used and Intellij's suggestion was incorrect due to the Groovy/Gradle SDK not being setup.

pjfanning (approved):
lgtm

PR #148, opened 2023-08-07
Add signing for release artifacts

mdedetrich:
I havent tried this out yet, mainly just used the instructions from https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signing_publications but I suspect we will find out if the signing works when we get there.

### DIFF
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -2,6 +2,7 @@ plugins {
   id 'groovy'
   id 'java-gradle-plugin'
   id 'maven-publish'
+  id 'signing'
   id 'io.github.gradle-nexus.publish-plugin' version '1.0.0'
 }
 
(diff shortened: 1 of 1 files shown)


mdedetrich:
Ill work on this ore tomorrow, seems like we are using such an old version of gradle that the way to setup signing is different.

gradle-plugin/build.gradle
```diff
@@ -81,6 +82,16 @@ nexusPublishing {
     }
 }
 
+Boolean isReleaseVersion = !version.toString().endsWith("SNAPSHOT")
+
+signing {
+    setRequired({
+        isReleaseVersion && gradle.taskGraph.hasTask("publish")
+    })
+    useGpgCmd()
```
mdedetrich, line 91: Gradle will use JVM signing, `useGpgCmd()` is set because people such as myself have their signing key stored on a key device (i.e. YubiKey) which doesn't work with JVM signing.

Its also consistent with sbt-gpg (which also uses gpg).

pjfanning (approved):
lgtm

PR #151, opened 2023-08-08
Add sbt-license-report

mdedetrich:
Adds sbt license report


pjfanning (approved):
lgtm

project/LicenseReport.scala
```diff
@@ -0,0 +1,48 @@
...
+object LicenseReport extends AutoPlugin {
+
+  override lazy val projectSettings = Seq(
+    licenseReportTypes := Seq(MarkDown),
+    licenseReportMakeHeader := (language => language.header1("License Report")),
+    licenseConfigurations := Set("compile", "test", "provided"),
+    licenseDepExclusions := {
+      case DepModuleInfo("org.apache.pekko", _, _) => true // Inter pekko project dependencies are pointless
+      case DepModuleInfo(_, "scala-library", _)    => true // Scala library is part of Scala language
+      case DepModuleInfo(_, "scala-reflect", _)    => true // Scala reflect is part of Scala language
```
He-Pin, line 32: Can those be a configurable set instead of three cases.

project/LicenseReport.scala
```diff
@@ -0,0 +1,48 @@
...
+    },
+    licenseOverrides := {
+      case dep: DepModuleInfo if dep.organization.startsWith("javax") => LicenseInfo.CDDL_GPL
```
He-Pin, line 35: Newer version  will starts.with Jakarta

scripts/link-validator.conf
```diff
@@ -24,7 +24,44 @@ site-link-validator {
...
+    "http://openjdk.java.net/legal/gplv2+ce.html",
+    "http://opensource.org/licenses/",
+    "http://www.opensource.org/licenses/",
+    "http://parboiled.org",
+    "http://pholser.github.com/jopt-simple",
+    "http://source.android.com/",
+    "http://www.apache.org/licenses",
+    "http://www.eclipse.org/legal/",
+    "http://www.gnu.org/licenses/",
+    "http://www.jcraft.com/jsch/",
```
He-Pin, line 59: Move to front,  get the list sorted.

He-Pin (approved):
Lgtm with  little comments

project/LicenseReport.scala
mdedetrich: Its probably better off to fix this by doing a master config in a hypothetical future `incubator-pekko-sbt-build` plugin

project/LicenseReport.scala
mdedetrich: The point of these license overrides is to fix invalid urls that the licenses point to, newer versions with Jakarta have valid links.

In other words this is deliberate, its only for older licenses that it needs to be fix.

scripts/link-validator.conf
mdedetrich: This list is temporary, will be removed when more functionality will get added to a forked link-validator

PR #152, opened 2023-08-08
Remove gradle plugin portal link in link-validator.conf

mdedetrich:
Isn't valid anymore since we don't publish to plugin portal


He-Pin (approved):
Lgtm

PR #155, opened 2023-08-09
get jar signing to work

pjfanning:
Release mangers will need to set up their `~/.gradle/gradle.properties` when creating a release candidate.

* https://docs.gradle.org/6.9.1/userguide/signing_plugin.html#sec:conditional_signing

They will need to add these props but correct the values to match their gpg setup.
```
signing.keyId=24875D73
signing.password=secret
signing.secretKeyRingFile=/Users/me/.gnupg/secring.gpg
```


mdedetrich (approved):
lgtm

PR #127, opened 2023-08-04
remove gradle-wrapper jars

pjfanning:
* copies code from Apache Kafka that allows us to remove the gradle-wrapper jars
* https://lists.apache.org/thread/33qbxl9ysw23gs4olp4gcxtp810sw8b4
* relates to #124

PR #156, opened 2023-08-09
remove gradlew bat file

pjfanning:
I meant to remove these in https://github.com/apache/incubator-pekko-grpc/pull/127

The problem is that gradlew.bat needs the gradle-wrapper.jar files to be pre-installed.

Windows users can install gradle.bat on their computers and use gradle.bat instead of gradlew.bat.

We don't have a Windows equivalent of the Apache Kafka derived gradlew script that works even when gradle-wrapper.jar files are missing.


mdedetrich (approved):
lgtm

PR #158, opened 2023-08-11
Rename sbt plugin artifact so it's consistent with other artifacts

mdedetrich:
Currently the artifact name for the sbt plugin is inconsistent with the other artifacts that pekko-grpc generates, i.e. currently the sbt artifact name is `sbt-pekko-grpc` where as all other artifacts start with `pekko-grpc-<REST>`. This even includes the other build tool plugins that pekko-grpc provides, i.e. `pekko-grpc-gradle-plugin`/`pekko-grpc-maven-plugin`

### DIFF
--- a/docs/src/main/paradox/release-notes/index.md
+++ b/docs/src/main/paradox/release-notes/index.md
@@ -30,6 +30,9 @@ We haven't had to fix any significant bugs that were in Akka gRPC 2.1.6.
   `mavenCentral()` to the `pluginManagement`'s `repositories` entry. See
   [Installation docs](https://pekko.apache.org/docs/pekko-grpc/current/buildtools/gradle.html#installation) for more
   info.
+* The naming convention of the Pekko gRPC sbt plugin has changed, i.e. whereas
+  with Akka the artifact was named `sbt-akka-grpc` with Pekko it's named
+  `pekko-grpc-sbt-plugin` so it's consistent with `pekko-grpc-gradle-plugin`/`pekko-grpc-maven-plugin`.
 
 ### Additions
 
--- a/build.sbt
+++ b/build.sbt
@@ -133,7 +133,7 @@ lazy val sbtPlugin = Project(id = "sbt-plugin", base = file("sbt-plugin"))
   .disablePlugins(MimaPlugin)
   .settings(Dependencies.sbtPlugin)
   .settings(
-    name := s"sbt-$pekkoPrefix",
+    name := s"$pekkoPrefix-sbt-plugin",
     /** And for scripted tests: */
     scriptedLaunchOpts += ("-Dproject.version=" + version.value),
     scriptedLaunchOpts ++= sys.props.collect { case (k @ "sbt.ivy.home", v) => s"-D$k=$v" }.toSeq,
--- a/sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/project/plugins.sbt
+++ b/sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/project/plugins.sbt
@@ -7,4 +7,4 @@
  * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
-addSbtPlugin("org.apache.pekko" % "sbt-pekko-grpc" % sys.props("project.version"))
+addSbtPlugin("org.apache.pekko" % "pekko-grpc-sbt-plugin" % sys.props("project.version"))
--- a/sbt-plugin/src/sbt-test/gen-java/02-server-reflection/project/plugins.sbt
+++ b/sbt-plugin/src/sbt-test/gen-java/02-server-reflection/project/plugins.sbt
@@ -7,4 +7,4 @@
  * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
-addSbtPlugin("org.apache.pekko" % "sbt-pekko-grpc" % sys.props("project.version"))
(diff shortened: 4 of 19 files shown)


pjfanning:
I don't think this is necessary. I'm still going to fight to get the RC1 approved.

mdedetrich:
> I don't think this is necessary. I'm still going to fight to get the RC1 approved.

I would prefer to get this done because we can't change it later easily without breaking users

pjfanning (changes requested):
-1 from me

We have a viable RC in progress and the name matches the akka one

mdedetrich:
> -1 from me
> 
> We have a viable RC in progress and the name matches the akka one

Im sorry but its more important to get the naming for the artifacts correct and consistent over a possible saving of a single day on an RC with only a negative vote.

docs/src/main/paradox/release-notes/index.md
```diff
@@ -30,6 +30,9 @@ We haven't had to fix any significant bugs that were in Akka gRPC 2.1.6.
   `mavenCentral()` to the `pluginManagement`'s `repositories` entry. See
   [Installation docs](https://pekko.apache.org/docs/pekko-grpc/current/buildtools/gradle.html#installation) for more
   info.
+* The naming convention of the Pekko gRPC sbt plugin has changed, i.e. where
+  as with Akka the artifact was named `sbt-akka-grpc` with Pekko its named
+  `pekko-grpc-sbt-plugin` so its consistent with `pekko-grpc-gradle-plugin`/`pekko-grpc-maven-plugin`.
```
pjfanning, line 35: should be `it's` twice - both cases are short for `it is`

pjfanning (changes requested):

docs/src/main/paradox/release-notes/index.md
mdedetrich: Fixed.

mdedetrich:
@pjfanning PR passing and the grammer fixes you have pointed out has been applied

pjfanning:
I'm still pondering this. I think we might need a vote on the dev list.

mdedetrich:
The arguments from my end should be clear, unlike all of our other artifacts which start with pekko (this can clearly be seen at https://repository.apache.org/content/groups/staging/org/apache/pekko/) currently the sbt-plugin is an exception. I also don't know if there is an ASF rule regarding artifact naming needing to start with `pekko`, but there is one with github repo's (i.e. all of our repos need to start with `incubating-pekko`)

In other words it sticks out with a giant sore thumb and since we are dealing with artifact naming it should be fixed before the first release to avoid unnecessary disruption later (this should be obvious). 

If you think its necessary then go ahead but I hardly see how this is that controversial.

gmethvin (approved):

pjfanning:
Up until today, all the Pekko artifacts have been named after the Akka ones that they are based on.

This artifact is based on https://central.sonatype.com/artifact/com.lightbend.akka.grpc/sbt-akka-grpc

Why should we break the pattern for this?

mdedetrich:
> Up until today, all the Pekko artifacts have been named after the Akka ones that they are based on.

No they haven't, the pekko-connectors for example is entirely different from the akka ones (which used to be called alpakka)

> Why should we break the pattern for this

I already provided the arguments before

pjfanning:
> > Up until today, all the Pekko artifacts have been named after the Akka ones that they are based on.
> 
> No they haven't, the pekko-connectors artifact naming for example is entirely different from the akka ones (which used to be called alpakka)
> 
> > Why should we break the pattern for this
> 
> I already provided the arguments before

The pekko connectors name change was agreed many many months. It wasn't decided on just as we attempted to do releases.

mdedetrich:
> Changing this name is going to cause major downstream disruption in builds that use pekko-grpc.

> These include builds include

> incubator-pekko-connectors
> incubator-pekko-samples

These are unreleased

> Still -1 from me

Then put it up on the mailing list, this PR already got an approval from someone else.

mdedetrich:
Also as far as I know, there was no real community decision to rename Alpakka to Connectors, at least I don't remember voting on it. It was just done (since Alpakka didn't make sense in the context of Pekko, it being a pun on Akka) and no one happened to complain about the name.

gmethvin:
I prefer the naming actually be consistent.

I guess one argument for keeping it the same as Akka is that it makes updating artifact names slightly easier when migrating a project to Pekko, but this is a very small one time cost.

pjfanning:
Ok. I'll remove my -1 since we need to go ahead and do an RC2. @mdedetrich can you fix up the other repos that I mentioned when we get a new grpc snapshot?

pjfanning (commented):
remove -1

mdedetrich:
> Ok. I'll remove my -1 since we need to go ahead and do an RC2. @mdedetrich can you fix up the other repos that I mentioned when we get a new grpc snapshot?

Will do, I'll generate new snapshots just after this merge and update downstream dependencies.

mdedetrich:
New version with artifact changes is 1.0.0-RC1-3-ae23c14d-SNAPSHOT

* pekko-connectors PR at https://github.com/apache/incubator-pekko-connectors/pull/217
* pekko-samples PR at https://github.com/apache/incubator-pekko-samples/pull/62

I have also updated the `Pekko 1.0.1 & Pekko HTTP 1.0.0 & Pekko Connectors Kafka 1.0.0` entry at https://cwiki.apache.org/confluence/display/PEKKO/Testing+with+Pekko+Snapshot+Jars

PR #162, opened 2023-08-11
add license details for proto file

pjfanning:
I'm not 100% sure but the file we inherited from akka-grpc looks very similar to a file I found on the web and it was checked in in February 2020 - a month before the Akka file.

https://github.com/cloudstateio/dotnet-support/blob/master/examples/Crdt.Example/proto_files/example/crdts/crdt-example.proto

https://github.com/akka/akka-grpc/commits/main/interop-tests/src/main/protobuf/crdt-example.proto

There is no source header on either file and the LICENSE file in the https://github.com/cloudstateio/dotnet-support/ project is a standard Apache v2.0 license with no specific copyright in it.

https://github.com/cloudstateio/dotnet-support/

PR #163, opened 2023-08-11
Delete crdt-example.proto

pjfanning:
I ran `sbt interop-tests/test` and this file wasn't needed

* see this description on #162 for why it is useful to get rid of this file


mdedetrich:
So I just had a quick glance at the code and I am getting the impression that this is meant to be an example file, in which case its fine to delete but maybe we should make a reference to the file in our documentation somewhere?

pjfanning:
* it is checked into a test project
* my impression is that it was meant to be part of the testing of a PR but then it wasn't used
* if we were to document where you find example protos - what page would it go on? If we have no such page, would we create a  page with just this proto?
* can't we add docs at any time?
* this file is a potential -1 in our release

PR #167, opened 2023-08-18
steps towards making gradle build reproducible

pjfanning:
https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives


nvollmar (approved):
lgtm

PR #168, opened 2023-08-20
upgrade protobuf-java to 3.21.12

pjfanning:
grpc lib uses a newer version of protobuf-java (3.21.1) then we set (3.20.3) but this has CVEs

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,7 +36,7 @@ object Dependencies {
     // If changing this, remember to update protoc plugin version to align in
     // maven-plugin/src/main/maven/plugin.xml and org.apache.pekko.grpc.sbt.PekkoGrpcPlugin
     val googleProtoc = "3.20.1" // checked synced by VersionSyncCheckPlugin
-    val googleProtobufJava = "3.20.3"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -36,7 +36,7 @@ object Dependencies {
     // If changing this, remember to update protoc plugin version to align in
     // maven-plugin/src/main/maven/plugin.xml and org.apache.pekko.grpc.sbt.PekkoGrpcPlugin
     val googleProtoc = "3.20.1" // checked synced by VersionSyncCheckPlugin
```
He-Pin, line 38: Can this two version be the same?

mdedetrich: I guess we should also use a version of protoc that doesn't have vulnerabilities but this is maybe for 1.1.x.

mdedetrich (approved):
lgtm

project/Dependencies.scala
pjfanning: the vulnerabilities are in protobuf-java, not in protoc

He-Pin (approved):
lgtm

Issue #171, opened 2023-09-03
upgrade grpc-protobuf due to CVEs


pjfanning:
See https://mvnrepository.com/artifact/org.apache.pekko/pekko-grpc-runtime_2.13/1.0.0

The newer versions of grpc use even newer versions of protobuf-java than we upgraded to in https://github.com/apache/incubator-pekko-grpc/pull/168 - so we will probably need to upgrade protobuf-java to an even newer version.

Edit: I tried newer grpc-protobuf versions but they caused compile problems. v1.53.0 is the oldest version of grpc-protobuf that has no CVEs and works without too many changes to pekko-grpc. It doesn't require an upgrade to pekko-grpc's protobuf-java dependency.

pjfanning:
https://github.com/apache/incubator-pekko-grpc/pull/172 merged

PR #172, opened 2023-09-03
upgrade grpc-protobuf due to CVEs

pjfanning:
see #171

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -163,6 +163,8 @@ lazy val interopTests = Project(id = "interop-tests", base = file("interop-tests
     ReflectiveCodeGen.generatedLanguages := Seq("Scala", "Java"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
     ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
+    // grpc 1.53.0 brings in extra unnecessary proto files that cause build issues
+    PB.generate / excludeFilter := new SimpleFileFilter(f => f.getAbsolutePath().startsWith("envoy")),
     PB.protocVersion := Dependencies.Versions.googleProtoc,
     // We need to be able to publish locally in order for sbt interopt tests to work
     // however this sbt project should not be published to an actual repository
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -163,6 +163,8 @@ lazy val interopTests = Project(id = "interop-tests", base = file("interop-tests
     ReflectiveCodeGen.generatedLanguages := Seq("Scala", "Java"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
     ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
+    // grpc 1.53.0 brings in extra unnecessary proto files that cause build issues
+    PB.generate / excludeFilter := new SimpleFileFilter(f => f.getAbsolutePath().contains("envoy/")),
```
He-Pin, line 167: Drop the `/`

mdedetrich (approved):
lgtm, but I would confirm with @He-Pin that this also works on windows.

build.sbt
pjfanning: made the change - I suspect there might be other issues with building this module on Windows but I think this PR should not now make it worse

build.sbt
```diff
@@ -163,6 +163,8 @@ lazy val interopTests = Project(id = "interop-tests", base = file("interop-tests
     ReflectiveCodeGen.generatedLanguages := Seq("Scala", "Java"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
     ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
+    // grpc 1.53.0 brings in extra unnecessary proto files that cause build issues
```
raboof, line 166: What kinds of issues? Should we create a GitHub issue for these and link it here?

raboof (approved):

build.sbt
pjfanning: just for reference, the build failures look like https://github.com/apache/incubator-pekko-grpc/actions/runs/6065657613/job/16455401558

The opencensus protos that cause the issues are imported by envoy protos. We don't seem to need these envoy protos. 

Since the issue only popped up after upgrading the io.grpc libs, I have to presume that the protos are being brought in by those jars.

pjfanning: the extra envoy and opencensus protos appear to have been added to grpc-interop-testing jar between the 1.48.1 and 1.53.0 releases

He-Pin (approved):
lgtm

PR #173, opened 2023-09-04
use newer version of guava due to CVE

pjfanning:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-2976

PR #175, opened 2023-09-05
gradle 7

pjfanning:
* may help with #173 - which is currently blocked due to odd issues loading guava jars using gradle
* see the comments in https://github.com/google/guava/issues/6617

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -153,7 +153,7 @@ jobs:
           cp .jvmopts-ci .jvmopts
           sbt codegen/publishM2 scalapb-protoc-plugin/publishM2 +runtime/publishM2
           cd gradle-plugin
-          ./gradlew clean publishToMavenLocal --console=plain --info --stacktrace
+          ./gradlew clean test publishToMavenLocal --console=plain --info --stacktrace
           find ~/.m2 | grep gradle
 
       - name: Test Gradle Java ${{ matrix.SCALA_VERSION }}
--- a/gradle-plugin/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle-plugin/gradle/wrapper/gradle-wrapper.properties
@@ -5,4 +5,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.2-bin.zip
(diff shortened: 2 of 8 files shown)


pjfanning:
This helps https://github.com/apache/incubator-pekko-grpc/pull/173/ - seems best still to merge this first and not to bundle the change in 173. I can rebase 173 after this is merged.

gradle-plugin/src/test/groovy/func/GradleCompatibilitySpec.groovy
```diff
@@ -42,7 +42,7 @@ class GradleCompatibilitySpec extends BaseSpec {
         then:
         result.task(":tasks").outcome == TaskOutcome.SUCCESS
         where:
-        gradleVersion << ["5.6", "5.6.4", "6.9.1"]
+        gradleVersion << ["5.6", "5.6.4", "7.6.2"]
```
raboof, line 45: this looks suspicious - does this mean it also tests "5.6" and "5.6.4"? Are these tests actually ran?

raboof (approved):
One thing to doublecheck, otherwise LGTM

gradle-plugin/src/test/groovy/func/GradleCompatibilitySpec.groovy
pjfanning: I'll have a look - it does look like these unit tests may not be part of the CI build

gradle-plugin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -5,4 +5,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.2-bin.zip
```
He-Pin, line 8: Could we extract the version to a variable? of the URL is stable.

pjfanning: there are millions of these files in github - Gradle have not provided a tidier mechanism - I don't want to reinvent Gradle - can we wait for Gradle to make this easier instead of me spending time on it?

pjfanning: is sbt any better, we have dozens of build.properties files checked in that all say `sbt.version=1.9.4`

.github/workflows/build-test.yml
```diff
@@ -153,7 +153,7 @@ jobs:
           cp .jvmopts-ci .jvmopts
           sbt codegen/publishM2 scalapb-protoc-plugin/publishM2 +runtime/publishM2
           cd gradle-plugin
-          ./gradlew clean publishToMavenLocal --console=plain --info --stacktrace
+          ./gradlew clean test publishToMavenLocal --console=plain --info --stacktrace
```
He-Pin, line 156: great!

He-Pin (approved):
lgtm

gradle-plugin/src/test/groovy/func/GradleCompatibilitySpec.groovy
He-Pin: As we updated to 7, maybe that should change to tests major versions  5 6 7

Issue #180, opened 2023-09-22
Publish the pekko-grpc-codegen package for Scala 2.13


brendo-m:
Can we have the codegen package published for 2.13 as well as 2.12 please? akka-grpc-codegen is published for 2.13 (albeit in a roundabout way) and this is a blocker for migrating to pekko-grpc for non-sbt build tools, e.g. for mill and in our case bazel.

More context:
* https://github.com/akka/akka-grpc/issues/949
* https://github.com/akka/akka-grpc/issues/1400

Thanks!

pjfanning:
New snapshot jars including ones for Scala 2.13 and Scala 3 have been published. See https://repository.apache.org/content/groups/snapshots/org/apache/pekko/

We may release a v1.0.1 in the next few weeks. No firm plans yet.

PR #181, opened 2023-09-22
support scala 2.13 and scala 3 build for codegen lib

pjfanning:
* upgrades twirl plugin so that we can support scala 3
* CodeGenerator needs to be different in Scala 2.12 (needs a nowarn but this causes newer Scala versions to not allow compilation)
* relates to #180


codegen/src/main/scala-2.13+/org/apache/pekko/grpc/gen/CodeGenerator.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  /** Generator name; example: `pekko-grpc-scala` */
+  def name: String
+
+  def run(request: CodeGeneratorRequest, logger: Logger): CodeGeneratorResponse
+
+  /** Takes Scala binary version and returns suggested dependency Seq */
+  def suggestedDependencies: ScalaBinaryVersion => Seq[Artifact]
+
+  def registerExtensions(registry: ExtensionRegistry): Unit = {}
```
He-Pin, line 35: Add sclaladoc too？

He-Pin (approved):
Lgtm

codegen/src/main/scala-2.13+/org/apache/pekko/grpc/gen/CodeGenerator.scala
pjfanning: the original code (still in use for scala 2.12 compile) does not have scaladoc so I don't want to make the 2 versions differ from each other

He-Pin: okay, but you can add doc in both place :)

pjfanning: I'm not a fan of hijacking PRs to make unrelated changes - if you want new scaladoc, create a new Issue or PR

PR #182, opened 2023-09-27
Switch to guava jre flavor

kciesielski:
https://github.com/softwaremill/tapir/pull/3124 is a PR which adds `pekko-grpc` support to Tapir. We noticed that `pekko-grpc-sbt-plugin` forces the android guava flavor:
```scala
"com.google.guava" % "guava" % "32.1.2-android" % "runtime"
```
It's described in a comment, that this, in effect, `forces a newer version than grpc-protobuf defaults too`.

The issue with Android flavor is that it overrides guava used by `sbt-scalajs`, which fails the `fastops` step with errors like:
```
[info] Fast optimizing /home/runner/work/tapir/tapir/integrations/cats/target/js-3/tapir-cats-test-fastopt
[error] java.lang.NoSuchMethodError: 'java.util.stream.Collector com.google.common.collect.ImmutableList.toImmutableList()'
[error] 	at org.openqa.selenium.remote.server.DefaultDriverFactory.<clinit>(DefaultDriverFactory.java:78)
```

Switching back to the `jre` flavor in Tapir seems to fix the issue. 
Can this be the default flavor used by pekko-grpc, since version number is locked to 32.1.2 anyway?

### DIFF
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -124,7 +124,7 @@ repositories {
 dependencies {
   implementation 'com.google.protobuf:protobuf-gradle-plugin:0.9.4'
   implementation 'commons-lang:commons-lang:2.6'
-  runtimeOnly 'com.google.guava:guava:32.1.2-android'
+  runtimeOnly 'com.google.guava:guava:32.1.2-jre'
   testImplementation 'junit:junit:4.13.2'
   testImplementation 'org.spockframework:spock-core:2.3-groovy-3.0'
 }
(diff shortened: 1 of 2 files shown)


kciesielski:
@pjfanning could you take a look at this?

He-Pin:
@pjfanning in a later pr, we do add it explicitly.

@kciesielski Excludes the android one in tapir will works too, I think it's wired to depend on an android favor, but that it's in the upstream.

kciesielski:
Sure, if you want to keep it consistent with the upstream, then I'll close this PR for now.

PR #189, opened 2023-11-10
re-enable mima checks

pjfanning:
based on how this is set in akka-grpc (at the time we split out pekko)


pjfanning:
@nvollmar thanks for reviewing all the MiMa PRs. This is the final one for now. With this, all the released modules will have MiMa checks enabled.

nvollmar (approved):
lgtm

PR #190, opened 2023-11-11
[DRAFT] io.grpc 1.59

pjfanning:
newer grpc version and protobuf-java upgraded to match what grpc 1.59.0 is built with


pjfanning:
superceded by #222

PR #197, opened 2023-12-25
[Netty] Fix client with TLS and custom SSL context

ihostage:
Fixes https://github.com/playframework/play-grpc/issues/468

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
@@ -184,6 +184,7 @@ object NettyClientUtils {
   private def createNettySslContext(javaSslContext: SSLContext): SslContext = {
     import io.grpc.netty.shaded.io.netty.handler.ssl.{
       ApplicationProtocolConfig,
+      ApplicationProtocolNames,
       ClientAuth,
       IdentityCipherSuiteFilter,
       JdkSslContext
@@ -195,8 +196,12 @@ object NettyClientUtils {
       /* boolean isClient */ true,
       /* Iterable<String> ciphers */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
       IdentityCipherSuiteFilter.INSTANCE,
-      /* ApplicationProtocolConfig apn */ ApplicationProtocolConfig.DISABLED, // use JDK default (null would also be acceptable, DISABLED config will select the NONE protocol and thus the JdkDefaultApplicationProtocolNegotiator)
-      ClientAuth.NONE, // server-only option, which is ignored as isClient=true (as indicated in constructor Javadoc)
+      new ApplicationProtocolConfig(
+        ApplicationProtocolConfig.Protocol.ALPN,
+        ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
+        ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
+        ApplicationProtocolNames.HTTP_2),
+      ClientAuth.OPTIONAL, // server-only option, which is ignored as isClient=true (as indicated in constructor Javadoc)
       /* String[] protocols */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
       /* boolean startTls */ false)
   }


pjfanning:
Is there any straightforward way to unit test this? I accept that this is a bug that needs to be fixed but ideally, we would have some regression test coverage.

ihostage:
> Is there any straightforward way to unit test this? I accept that this is a bug that needs to be fixed but ideally, we would have some regression test coverage.

I'll try adopt our failed tests from Play... Give me a time 😉

pjfanning:
> > Is there any straightforward way to unit test this? I accept that this is a bug that needs to be fixed but ideally, we would have some regression test coverage.
> 
> I'll try adopt our failed tests from Play... Give me a time 😉

Thanks

pjfanning:
Ok. I'll merge this but we'll need to look into adding tests before we do a release.

pjfanning (approved):
lgtm - tests pending

pjfanning:
this is going to have to be reverted - it looks like the code was taken from akka-grpc - this is unacceptable

ihostage:
> this is going to have to be reverted - it looks like the code was taken from akka-grpc - this is unacceptable

Sorry @pjfanning, It looks like some kind of nightmare to me 😱 

I'll try to describe what this history was on my side. 

1. As a co-maintainer @Playframework, I found time to work on Play gRPC that haven't had our love in long time (by many reasons and our focus to more priority issues)
2. I caught "_ALPN must be enabled and list HTTP/2 as a supported protocol_" error while was working on this update https://github.com/playframework/play-grpc/pull/407
3. Like you in https://github.com/apache/incubator-pekko-grpc/pull/200, I also analysed a diff between `2.1.5...2.1.6` and found a changes in https://github.com/apache/incubator-pekko-grpc/commit/b0ff79ecc4382b25bad4e041df6cdbaa3b305343 that were a root cause a problem in our tests.
4. I didn't want revert these changes and back to using reflection 🤮 And I jumped to source `GrpcSslContexts` that was used before to find a solution. 
5. Shortly, my route was next:
[GrpcSslContexts.configure(SslContextBuilder builder, SslProvider provider)](https://github.com/grpc/grpc-java/blob/6e2e18bb728793df32b2ba195a954ad380e546de/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L155)
⬇️ 
[return configure(builder, jdkProvider)](https://github.com/grpc/grpc-java/blob/6e2e18bb728793df32b2ba195a954ad380e546de/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L164)
⬇️ 
[configure(SslContextBuilder builder, Provider jdkProvider)](https://github.com/grpc/grpc-java/blob/6e2e18bb728793df32b2ba195a954ad380e546de/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L189)
⬇️ 
[apc = ALPN](https://github.com/grpc/grpc-java/blob/6e2e18bb728793df32b2ba195a954ad380e546de/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L212)
⬇️ 
[private static final ApplicationProtocolConfig ALPN = ...](https://github.com/grpc/grpc-java/blob/6e2e18bb728793df32b2ba195a954ad380e546de/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L66-L70)
⬇️
Got it! 
6. A couldn't use `ALPN` variable directly because it's private 😞  That why I just copied the declaration of this variable and pass it instead of `ApplicationProtocolConfig.DISABLED`. Yes, I copied this line, but from gRPC-Java that is under APL 2 not from Akka gRPC. Of course, also needed to change a private variable `NEXT_PROTOCOL_VERSIONS` to `ApplicationProtocolNames.HTTP_2` but it was a trivial to find a public suitable variable. 
7. I published locally a patched version Akka gRPC. **It was anought**, Play gRPC tests was green locally and I was happy.
8. Because Play 2.9.x based on Akka and we support this version, I wanted to create an issue and PR for Akka gRPC to fix this problem and backport this fix to OSS `2.1.x` version. Then I found that it's already fixed and Akka gRPC has a related issue 🤦‍♂️ 
8. I asked my friends (maybe it's a one way friendship but I consider them my friends) from LB backport their fix to `2.1.x` and created PR with **their** (`cherry-pick` helped me and it's important for the story) changes to simplify as much as possible their works to release a new OSS version. (You can find these discussions and related PR in Akka gRPC, I don't want to link their here)
9. Also important to note, that as you can see I tried to fix a problem for Akka firstly when we haven't had a Play gRPC version is migrated to Pekko. After this adventure to fix this problem for Play 3 (Play gRPC `0.12.x`), I ported changes to Pekko gRPC repository that was applied to Akka gRPC. And in this step I made a mistake (don't remove unnecessary `ClientAuth.OPTIONAL` from my changes) and sorry for that 😢 Maybe I made a mistake on _step 8_, I already don't know, but I can't imagine how I can miss this step if it is in my nature 🤷‍♂️ 

**Summary:**
So, now you know how `ClientAuth.NONE -> ClientAuth.OPTIONAL` became a part of my PR. But I insist that other part is legal. And this whole situation is terrible to me, because I don't understand yet what my development process on Play should be without touching Akka source code (include the latest sometimes). Of course, as a solution I just may do nothing for Pekko but I dislike this way 😞 

**Just thinking out loud**
Instead of improve a software we should thinking about legal every line changes... Ouch, it's incredible, folks, really incredible...

He-Pin:
@ihostage I trust you , but it's really hard to keep this safe.

Btw, I'm trying to sync some fix to akka/akka too, as: https://github.com/akka/akka/pull/32269

Akka is great, but for me I can't use it at work anymore, it's not decided by me.

And for the actual fix, we can definitely come up with a better fix.

pjfanning:
thanks @ihostage for the detailed explanation

Unfortunately, we're in a position where companies will sue over a single line of code. Some companies buy up software and patents with a view to litigating.

PR #199, opened 2023-12-25
Revert "[Netty] Fix client with TLS and custom SSL context"

pjfanning:
Reverts apache/incubator-pekko-grpc#197

the OP of apache/incubator-pekko-grpc#197 copied license incompatible code from akka-grpc - completely unacceptable

the code might be short but the OP kept the comments exactly the same as the akka comments making it obvious that the code is not his

Pekko contributors - anyone else who reads the akka-grpc code will become unavailable to review or contribute to fixing the original issue

We need a clean room fix.


samueleresca (approved):
On mobile, approving since it is a revert of previous changes

PR #200, opened 2023-12-25
revert grpc breaking netty tls change

pjfanning:
see #197 and #199 

the akka-grpc 2.1.6 breakage was traced back to this OSS akka-grpc 2.1.6 change (https://github.com/akka/akka-grpc/pull/1649) and reverted here

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
@@ -182,10 +182,7 @@ object NettyClientUtils {
    */
   @InternalApi
   private def createNettySslContext(javaSslContext: SSLContext): SslContext = {
-    import io.grpc.netty.shaded.io.netty.handler.ssl.{
-      JdkSslContext,
-      SslProvider
-    }
+    import io.grpc.netty.shaded.io.netty.handler.ssl.{ JdkSslContext, SslProvider }
     import java.lang.reflect.Field
 
     // This is a hack for situations where the SSLContext is given.
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
@@ -183,22 +183,25 @@ object NettyClientUtils {
   @InternalApi
   private def createNettySslContext(javaSslContext: SSLContext): SslContext = {
     import io.grpc.netty.shaded.io.netty.handler.ssl.{
-      ApplicationProtocolConfig,
-      ClientAuth,
-      IdentityCipherSuiteFilter,
-      JdkSslContext
+      JdkSslContext,
+      SslProvider
     }
-    // See
-    // https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java#L229-L309
-    new JdkSslContext(
-      javaSslContext,
-      /* boolean isClient */ true,
-      /* Iterable<String> ciphers */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
-      IdentityCipherSuiteFilter.INSTANCE,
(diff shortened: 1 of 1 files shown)


pjfanning:
adding @mkurz too 

Since we are now in license hell due to #197, I think the best starting place is to revert the original change that caused the issue (that was made before Pekko forked and we are safe to reference that change since it is an OSS friendly chane).

We can build on top of this but this might a starting point.

runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
```diff
@@ -183,22 +183,25 @@ object NettyClientUtils {
...
+    // Create a Netty JdkSslContext object with all the correct ciphers, protocol settings, etc initialized.
+    val nettySslContext: JdkSslContext =
+      GrpcSslContexts.configure(GrpcSslContexts.forClient, SslProvider.JDK).build.asInstanceOf[JdkSslContext]
+
+    // Patch the SSLContext value inside the JdkSslContext object
+    val nettySslContextField: Field = classOf[JdkSslContext].getDeclaredField("sslContext")
+    nettySslContextField.setAccessible(true)
+    nettySslContextField.set(nettySslContext, javaSslContext)
+
+    nettySslContext
```
He-Pin, line 201: so this is leveraging the netty and based on a special impl too, future netty can do some kind of refactoring, and then this will fail again.

pjfanning: it will work for now and we can build a new fix - code is a living thing, today's fixes can be improved later on

we have broken code today and #197 makes it hard for us to do a better fix than this in the short term

He-Pin (approved):
lgtm

runtime/src/main/scala/org/apache/pekko/grpc/internal/NettyClientUtils.scala
He-Pin: Nice find, I have not check it:) maybe someone will do some tweets on X then.
And for the fix, I think that should not be hard. the tests suit maybe a little complex which need some keys setup.
Before we come up a fix, we should provide a test about it first.

He-Pin: Another issue is, can't we just do a quick reset and git push -f to drop that commit?

pjfanning: > Another issue is, can't we just do a quick reset and git push -f to drop that commit?

I'm sure if removing the #197 commit from git history is required but if it the consensus is to try to hide it altogether, we can go that way.

He-Pin: We should just drop it completly I think, otherwise the result commit log contains incompatible code.

pjfanning: The thing is that the git branch is protected - it will need to be unprotected to allow commits to be removed. Anyone who has a local git branch with the commit would also need to clean their set up afterwards. So forcing a removal of the commit is pretty drastic. It is not clear that it is absolutely needed.

He-Pin: I know, but more safer way I think.

mdedetrich: I am against unprotecting the git branch, it may have been acceptable right at the time of forking pekko but at this point there are way too many people that have this git branch checked out and over-writing the `main` branch basically forces everyone who has git cloned this repo to re-pull the changes, so even if we do remove the commit from `main` unless people wipe their repo they will still have the commit.

Thing is, I don't know what the ASF policy on this is, if I had a say in this I would add the commit (and the revert commit) to `.git-ignore-blame-revs` so that its more hidden but realistically I don't know what more can be done.

pjfanning:
this seems to fix the play-grpc issue - https://github.com/playframework/play-grpc/issues/468#issuecomment-1869472024

mkurz:
Thanks for fixing this over the holidays! I see you just released v1.0.2, so I released play-grpc 0.12.1 as well:
- https://github.com/playframework/play-grpc/releases/tag/0.12.1

PR #203, opened 2023-12-26
use sbt-pekko-build to allow testing with pekko/pekko-http snapshots

pjfanning:
need to be able to test with different builds


He-Pin (approved):
Lgtm

laglangyue:
sbt-pekko-build is new build tool,it should be tested in different machine, linux/mac/win.

PR #205, opened 2023-12-27
test with sslcontext

pjfanning:
The new tests in GrpcInteropIoWithPekkoNettyScalaWithSslContextSpec fail without the fix in #199

### DIFF
--- a/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoGrpcScalaClientTester.scala
+++ b/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoGrpcScalaClientTester.scala
@@ -13,8 +13,6 @@
 
 package org.apache.pekko.grpc.interop
 
-import java.io.InputStream
-
 import org.apache.pekko
 import pekko.actor.ActorSystem
 import pekko.grpc.{ GrpcClientSettings, GrpcResponseMetadata, SSLContextUtils }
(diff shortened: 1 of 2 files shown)


interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoGrpcScalaClientTester.scala
```diff
@@ -41,21 +41,29 @@ import scala.util.control.NoStackTrace
  * The same implementation is also be found as part of the 'scripted' tests at
  * /sbt-plugin/src/sbt-test/gen-scala-server/00-interop/src/test/scala/org/apache/pekko/grpc/PekkoGrpcClientTester.scala
  */
-class PekkoGrpcScalaClientTester(val settings: Settings, backend: String)(implicit system: ActorSystem)
-    extends ClientTester {
+class PekkoGrpcScalaClientTester(val settings: Settings, backend: String, testWithSslContext: Boolean)(
```
He-Pin, line 44: maybe better with a trait
 SslContextProvider, and the false can be expressed by returning None.

pjfanning: It is test code. And the SSL context is created inside this class. Not outside.

pjfanning: When we have a use case for creating the SSL context outside - we can change then. Seems bad to overengineer test code

He-Pin (approved):
Lgtm

PR #213, opened 2024-01-24
Update sbt-pekko-build to 0.3.1

mdedetrich:
Updates sbt-pekko-build to 0.3.1 which brings in the pom.xml `name` field fix


gradle-plugin/build.gradle
```diff
@@ -44,7 +44,7 @@ gradlePlugin {
 publishing {
     publications.withType(MavenPublication) {
         pom {
-            name = "pekko-grpc-gradle-plugin"
+            name = "Apache Pekko Grpc Gradle Plugin"
```
He-Pin, line 47: There is a maven plugin too

mdedetrich: > There is a maven plugin too

Yes but the maven plugin is generated by sbt so the name will also be fixed by the updated sbt-apache-sonatype

He-Pin (approved):
Lgtm

PR #215, opened 2024-01-24
Use lazy val in build for zero args defs plus sbt dsl

mdedetrich:
Should improve the sbt build speed a little bit by avoiding unnecessary computation


He-Pin (approved):
Lgtm

PR #217, opened 2024-01-24
Remove redundant license info

mdedetrich:
This is redundant because sbt-license-info automatically generates a report with all of the license information from the dependency tree


pjfanning (approved):
lgtm

PR #224, opened 2024-02-07
mima comparison failing

pjfanning:
issue is that previousStableVersion only looks at git tags and we have a v1.1.0-M0 git tag but no release of that name

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -98,7 +98,7 @@ lazy val runtime = Project(id = "runtime", base = file("runtime"))
   .settings(
     name := pekkoGrpcRuntimeName,
     mimaFailOnNoPrevious := true,
-    mimaPreviousArtifacts :=  Set(
+    mimaPreviousArtifacts := Set(
       organization.value %% "pekko-grpc-runtime" % mimaCompareVersion),
     AutomaticModuleName.settings("pekko.grpc.runtime"),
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
Lgtm

raboof (approved):
Good to get unstuck now.

> we have a v1.1.0-M0 git tag but no release of that name

Do we expect this to be an exceptional situation or something that happens regularly? If the former, we should add an issue to remind us to remove the workaround after the 1.1.0 release. If the latter, we should add updating this value to the release procedure.

pjfanning:
I raised https://github.com/sbt/sbt-dynver/issues/283

In theory, we could submit such a fix ourselves and if it isn't merged, we can add it to one of our own sbt plugins.

mdedetrich:
> Do we expect this to be an exceptional situation or something that happens regularly? If the former, we should add an issue to remind us to remove the workaround after the 1.1.0 release. If the latter, we should add updating this value to the release procedure.

This is normal, we have to create a `M0` tag when we bump a major/minor so that nightly snapshots start using that new bumped major/minor version.

i.e. if we are at 1.1.x and we decide to bump the minor version we have to create a 1.2.0-M0 tag along with a branch so that nightly job will create versions starting with 1.2.0-x

Issue #228, opened 2024-02-11
add new compatibility test


pjfanning:
https://github.com/apache/incubator-pekko-grpc/tree/1.0.x/sbt-plugin/src/sbt-test/gen-scala-server/06-compatibility-plugins

this test doesn't work with newer versions of protobuf - see https://github.com/protocolbuffers/protobuf-javascript/issues/127

the idea is to remove it to allow us to proceed with io.grpc and protobuf upgrades

this issue is about replacing it

raboof:
switched to go

PR #229, opened 2024-02-11
remove compatibility test

pjfanning:
this test blocks #222 and it isn't a core feature

#228 is about writing a new test to replace this


raboof:
Looking at https://github.com/protocolbuffers/protobuf-javascript/issues/127#issuecomment-1542288860 are we sure it's impossible to make this work with newer versions of protobuf?

pjfanning:
@raboof to flip the question - is it worth blocking the upgrade of the grpc and protobuf libs until someone works out how to fix this test? https://github.com/apache/incubator-pekko-grpc/issues/228 is there to remind us that we need to add this or an equivalent test back.

raboof:
> #228 is there to remind us that we need to add this or an equivalent test back.

TBH I'm not optimistic that that would actually happen when there's no pressing need anymore ;)

> @raboof to flip the question - is it worth blocking the upgrade of the grpc and protobuf libs until someone works out how to fix this test? 

It seems we are fairly close... let's give it a week?

raboof:
superseded by #222

PR #248, opened 2024-03-22
recent failures in link-validator

pjfanning:
see https://github.com/apache/pekko-grpc/actions/runs/8390727921

we are at risk of license report issues all the time


samueleresca (approved):

raboof (approved):
I wonder if we can/should exclude license-report.html from the link validator wholesale

PR #256, opened 2024-04-05
Pin Scala 3.3 LTS in scala steward

mdedetrich:
I noticed in Scala OS projects that Scala Steward has been bumping 3.3.x to 3.4.x


mdedetrich:
@pjfanning Can we hold merging these PR's? I am currently trying to confirm that it actually works and we are blocked by https://github.com/apache/pekko/issues/1253

PR #278, opened 2024-04-06
upgrade protobuf/protoc to 3.25.3

pjfanning:
scala-steward doesn't upgrade protoc

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,8 +36,8 @@ object Dependencies {
     // Even referenced explicitly in the sbt-plugin's sbt-tests
     // If changing this, remember to update protoc plugin version to align in
     // maven-plugin/src/main/maven/plugin.xml and org.apache.pekko.grpc.sbt.PekkoGrpcPlugin
-    val googleProtoc = "3.24.0" // checked synced by VersionSyncCheckPlugin
-    val googleProtobufJava = "3.24.4"
+    val googleProtoc = "3.25.3" // checked synced by VersionSyncCheckPlugin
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -36,8 +36,8 @@ object Dependencies {
     // Even referenced explicitly in the sbt-plugin's sbt-tests
     // If changing this, remember to update protoc plugin version to align in
     // maven-plugin/src/main/maven/plugin.xml and org.apache.pekko.grpc.sbt.PekkoGrpcPlugin
-    val googleProtoc = "3.24.0" // checked synced by VersionSyncCheckPlugin
-    val googleProtobufJava = "3.24.4"
+    val googleProtoc = "3.25.3" // checked synced by VersionSyncCheckPlugin
+    val googleProtobufJava = "3.25.3"
```
raboof, line 40: can we drop `googleProtobufJava` now that they're in sync again?

raboof (approved):
one comment, but LGTM with or without it

Roiocam (approved):
LGTM

He-Pin (approved):
Lgtm

project/Dependencies.scala
pjfanning: I've set one to equal the other.
I think it is useful to have 3 props as it has gotten us out of jail before.
ScalaSteward is not going to generate correct PRs for us because the values also appear in Grade and Maven files too.

pjfanning: I did try to edit on my phone but edit seems to have failed. I will try again later.

PR #279, opened 2024-04-06
remove guava upgrade hack

pjfanning:
the old guava version hack was down to Pekko-gRPC 1.0 and our attempt to use an out of date io.grpc jar but to use a newer guava jar because of a CVE in the old guava version

We now use up to date io.gprc jar so it will have a newer guava dependency.

See dependencies in https://mvnrepository.com/artifact/io.grpc/grpc-protobuf/1.62.2


samueleresca (approved):

raboof (approved):
nice

PR #287, opened 2024-04-09
Refactoring some code of Codec

laglangyue:
## motivation
 Removed an unnecessary Set[CodeName] in Codec

## result 
clean up

## Test
test locally


raboof (approved):
nice

PR #297, opened 2024-04-14
chore:fix scalafmt HeadersSpec error because illegal unicode

laglangyue:
## motivation
can't execute sbt scalafmtAll
```
invalid unicode surrogate pair
[error]         ("my favorite character is \ud801", "my favorite character is ?")
```
according to test in grpc-java, we should use 0xD801.toChar
https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/test/java/io/grpc/StatusTest.java#L82-L92

### DIFF
--- a/runtime/src/test/scala/org/apache/pekko/grpc/scaladsl/headers/HeadersSpec.scala
+++ b/runtime/src/test/scala/org/apache/pekko/grpc/scaladsl/headers/HeadersSpec.scala
@@ -36,9 +36,9 @@ class HeadersSpec extends AnyWordSpec with Matchers {
         ("my favorite character is %", "my favorite character is %25"),
         ("my favorite character is 𐀁", "my favorite character is %F0%90%80%81"),
         // \ud801 is a high surrogate, a lone surrogate character is getting decoded as ? with UTF-8
-        ("my favorite character is \ud801", "my favorite character is ?"),
+        (s"my favorite character is ${0xD801.toChar}", "my favorite character is ?"),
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
Lgtm thanks

raboof:
huh, I could've sworn this syntax was allowed in Scala, but indeed I don't see it in https://scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#string-literals

laglangyue:
@raboof 
Here is a timeline:
I started to discover the problem, and I thought it was a problem with scalafmt, but members of scalafmt thought it was an abandoned unicode, which is actually very rare. This symbol has a universal unicode
Then I submitted a PR.
Finally, I argued with scalafmt that they are compatible with this Unicode form and will not throw exceptions

laglangyue:
Both writing methods are okay in Scala, and grpc-java uses `(char) "xxx"`. This pr will not affect the our test.
here is issue
https://github.com/scalameta/scalameta/issues/3690
Because this issue occurs after this PR,  So I didn't give any feedback on what happened afterwards to the Pekko community

raboof:
> both writing methods are okay in Scala

I agree the compiler currently appears to supports it, but I don't see anything in the language specification that says it must - so perhaps it would be best not to make assumptions? Or did I miss the part of the specification that covers this case?

> https://github.com/scalameta/scalameta/issues/3690

Thanks for bringing the issue up there and adding the reference here!

PR #230, opened 2024-02-24
make sure trailers are present in StatusRuntimeException

jtjeferreira:
When using the pekko-http backend and a `StatusRuntimeException` is returned it lacks metadata. 

This PR fixes the case for both headers and trailers

The issue was initially reported in https://github.com/akka/akka-grpc/issues/1814 and I fixed it myself in akka-grpc

PR #304, opened 2024-04-19
fix binary trailer decoding

jtjeferreira:
In #230 I introduced a bug in the trailer decoding where I didn't take into consideration that there are "ASCII headers" and "binary headers"


pjfanning (approved):
lgtm - thanks

PR #309, opened 2024-05-20
use ByteStringInputStream

pjfanning:
follow up to #308

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/ByteStringInputStream.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/ByteStringInputStream.scala
@@ -14,14 +14,10 @@
 package org.apache.pekko.grpc.internal
 
 import java.io.{ ByteArrayInputStream, InputStream, SequenceInputStream }
-import java.lang.invoke.{ MethodHandles, MethodType }
-
-import scala.util.Try
 
 import org.apache.pekko
 import pekko.annotation.InternalApi
 import pekko.util.ByteString
-import pekko.util.ByteString.ByteString1C
 import pekko.util.ByteString.ByteStrings
 import pekko.util.ccompat.JavaConverters._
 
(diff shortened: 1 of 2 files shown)


He-Pin (approved):
Lgtm

runtime/src/main/scala/org/apache/pekko/grpc/internal/ByteStringInputStream.scala
```diff
@@ -0,0 +1,58 @@
...
+      }
+  }
+
+  private def legacyConvert(bs: ByteString): InputStream = bs match {
+    case cs: ByteString1C =>
+      getInputStreamUnsafe(cs)
+    case _ =>
+      // NOTE: We actually measured recently, and compact + use array was pretty good usually
+      legacyConvert(bs.compact)
+  }
```
jtjeferreira, line 53: Does the note apply here? Cant we use something like this?

```suggestion
  private def legacyConvert(bs: ByteString): InputStream = bs match {
    case bytestrings: ByteStrings =>
      new SequenceInputStream(bytestrings.iterator.map(_.asInputStream).asJavaEnumeration)
    case _ =>
      new ByteArrayInputStream(bs.toArrayUnsafe())
  }
```

jtjeferreira: actually

```
def legacyConvert(bs: ByteString): InputStream = bs match {
      case bss: ByteStrings =>
        new SequenceInputStream(bss.bytestrings.iterator.map(legacyConvert).asJavaEnumeration)
      case _ =>
        new ByteArrayInputStream(bs.toArrayUnsafe())
    }
```

pjfanning: That does look better. I'll do some testing with later.

runtime/src/main/scala/org/apache/pekko/grpc/internal/ByteStringInputStream.scala
```diff
@@ -0,0 +1,59 @@
...
+  def apply(bs: ByteString): InputStream = bs match {
+    case cs: ByteString1C =>
+      getInputStreamUnsafe(cs)
+    case _ =>
+      if (byteStringInputStreamMethodTypeOpt.isDefined) {
+        byteStringInputStreamMethodTypeOpt.get.invoke(bs).asInstanceOf[InputStream]
+      } else {
+        legacyConvert(bs)
+      }
+  }
```
jtjeferreira, line 47: and what about:

```suggestion
  def apply(bs: ByteString): InputStream = {
    if (byteStringInputStreamMethodTypeOpt.isDefined) {
      byteStringInputStreamMethodTypeOpt.get.invoke(bs).asInstanceOf[InputStream]
    } else {
      legacyConvert(bs)
    }
  }
```

jtjeferreira: another improvement could be the use of `null` instead of `None` in `byteStringInputStreamMethodTypeOpt`

pjfanning: I'll check it again but when I was benchmarking the changes before - having the ByteString1C case seemed to help performance. I can't track down the benchmarks that I used, I may need to write them again.

pjfanning: actually - the benchmark code that I used a few weeks ago is at https://github.com/pjfanning/pekko-test/

pjfanning: With your optimised version of legacyConvert, we may not need the MethodHandle. That code is basically what adInputStream does anyway.

pjfanning: I checked in the simpler version. It seems good enough and Pekko 1.0 users benefit with this version.

pjfanning:
@He-Pin I reworked the code. Are you still happy with this change?

pjfanning:
@jtjeferreira It isn't strictly necessary but could you consider supplying an iCLA (if you haven't done so already)?

https://www.apache.org/licenses/contributor-agreements.html

laglangyue:
lgtm

nvollmar (approved):
lgtm

PR #321, opened 2024-05-29
chore: bump sbt to 1.10.0

He-Pin:
Motivation:
Bump sbt to 1.10.0

### DIFF
--- a/benchmark-java/project/build.properties
+++ b/benchmark-java/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.9.9
+sbt.version=1.10.0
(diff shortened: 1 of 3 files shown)


pjfanning (changes requested):
there are 2 other mentions of 1.9.9 in our source files - this is not the only one

pjfanning:
I have https://github.com/apache/pekko-grpc/pull/322

Updating Scala manually fixed the Scala Steward job. Updating Scala is harder than updating a lib because our CI yml files have Scala versions listed in them. That seems to be the permission that caused us the issue (the permission to update the CI yml files)

PR #357, opened 2024-08-11
non-https links now in license page

pjfanning:
https://github.com/apache/pekko-grpc/actions/runs/10341534635/job/28623448570?pr=354

We've got no control over what goes in that page.


laglangyue:
lgtm

raboof (approved):
Longer-term it would be good to do something like https://github.com/apache/pekko/pull/1381 , but this seems to get confused by the relative links somehow, so this is a good short-term step.

PR #362, opened 2024-09-14
in CI, don't specify full scala version

pjfanning:
* makes CI maintenance easier
* the sbt files have the full scala versions specified


pjfanning:
looks like the build needs full scala versions

PR #369, opened 2024-09-22
protobuf 4.28.2

pjfanning:
apparently ok - as grpc-protobuf is expected to work with protobuf 4 - recent comments in https://github.com/grpc/grpc-java/issues/11015


pjfanning:
looks like we need a new scalapb release - https://github.com/scalapb/ScalaPB/issues/1737

PR #381, opened 2024-10-03
scala 2.13.15 and 3.3.4

pjfanning:
seems ok to upgrade now

build issues are just in link-validator and those issues are explainable (gnu.org is offline and lightbemd.com page works but link-validator seems to be having issues with it in a few builds)


raboof (approved):
the link validator failure seems to be a https://www.gnu.org/ outage

PR #382, opened 2024-10-08
pekko 1.1.2

pjfanning:


raboof (approved):
why are we manually adding these instead of leaving this up to scala-steward again? was there some technical issue?

pjfanning:
> why are we manually adding these instead of leaving this up to scala-steward again? was there some technical issue?

ScalaSteward has no inling of how PekkoCoreDependency.scala and sbt-pekko-build works. It has attempted to produce PRs in some repos that were very broken.

raboof:
> > why are we manually adding these instead of leaving this up to scala-steward again? was there some technical issue?
> 
> ScalaSteward has no inkling of how PekkoCoreDependency.scala and sbt-pekko-build works. It has attempted to produce PRs in some repos that were very broken.
> 
> Example PR where ScalaSteward became very confused and suggested PRs that updated the wrong versions for the wrong modules.
> 
> https://github.com/apache/pekko-projection/pull/203/files

lol that was indeed less smart than I expected :laughing:

PR #391, opened 2024-10-23
Merge grpc headers and trailers in case of failure

NavidJalali:
Attempt to fix https://github.com/apache/pekko-grpc/issues/392

### DIFF
--- a/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoHttpServerProviderScala.scala
+++ b/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoHttpServerProviderScala.scala
@@ -120,7 +120,8 @@ object PekkoHttpServerProviderScala extends PekkoHttpServerProvider with Directi
         case _ =>
           response
             .attribute(AttributeKeys.trailer)
-            .map(trailer => Trailer(f(trailer.headers.map((RawHeader.apply _).tupled))))
+            .map(trailer => f(trailer.headers.map((RawHeader.apply _).tupled)))
+            .flatMap(headers => if (headers.isEmpty) None else Some(Trailer(headers)))
             .fold(response)(response.addAttribute(AttributeKeys.trailer, _))
       })
 }
--- a/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoHttpServerProviderScala.scala
+++ b/interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoHttpServerProviderScala.scala
@@ -118,11 +118,9 @@ object PekkoHttpServerProviderScala extends PekkoHttpServerProvider with Directi
                   HttpEntity.LastChunk(last.extension, f(last.trailer))
               }))
         case _ =>
-          val origTrailers = response
+          response
             .attribute(AttributeKeys.trailer)
-            .map(_.headers)
-            .getOrElse(Vector.empty)
-            .map(e => RawHeader(e._1, e._2))
-          response.addAttribute(AttributeKeys.trailer, Trailer(f(origTrailers)))
+            .map(trailer => Trailer(f(trailer.headers.map((RawHeader.apply _).tupled))))
+            .fold(response)(response.addAttribute(AttributeKeys.trailer, _))
       })
 }
--- a/interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcExceptionDefaultHandleSpec.scala
+++ b/interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcExceptionDefaultHandleSpec.scala
@@ -17,7 +17,7 @@ import org.apache.pekko
 import pekko.actor.ActorSystem
 import pekko.grpc.internal.{ GrpcProtocolNative, GrpcRequestHelpers, Identity }
 import pekko.grpc.scaladsl.headers.`Status`
-import pekko.http.scaladsl.model.{ HttpEntity, HttpRequest, HttpResponse }
+import pekko.http.scaladsl.model._
 import pekko.http.scaladsl.model.HttpEntity.{ Chunked, LastChunk, Strict }
 import pekko.grpc.GrpcProtocol
 import pekko.stream.scaladsl.{ Sink, Source }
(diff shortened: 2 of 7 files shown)


pjfanning:
A lot of unit tests are broken by this change.

NavidJalali:
@pjfanning yes, I am investigating them now. I didn't run the tests when I was checking if it fixed my issue.

NavidJalali:
Ran into an unrelated issue:
https://github.com/apache/pekko-http/pull/622

raboof:
> Ran into an unrelated issue: [apache/pekko-http#622](https://github.com/apache/pekko-http/pull/622)

That's strange, where does that Trailer without headers get created?

pjfanning:
I would not favour merging this. It ties us to an unreleased version of pekko-http - and I think there is an argument that the real issue is in the AWS code.

raboof:
> I would not favour merging this. It ties us to an unreleased version of pekko-http - and I think there is an argument that the real issue is in the AWS code.

I agree we should not merge this as long as it depends on an unreleased version of Pekko HTTP. The PR is still in 'Draft' mode, so that tracks ;).

I agree there's a case to be made that the 'real issue' of #392 is in the AWS implementation. That said, responding with a `Strict` response for errors like a 404 instead of streaming (which is the core of this PR) seems like a reasonable optimization regardless.

It might still be possible to adapt this PR to avoid creating the situation where a `Trailers` object without any headers is created, so it can go back to using the regular released Pekko HTTP version. In any case it'll still take a couple of rounds of improvements to make this change nice and ready to merge, but I'm optimistic about the chances of getting there.

raboof:
@NavidJalali thanks for working on this, let me know if you'd like me to review your work-in-progress!

pjfanning:
For practical reasons, it would be good to change this PR to avoid creating Trailers with no headers.

NavidJalali:
> > Ran into an unrelated issue: [apache/pekko-http#622](https://github.com/apache/pekko-http/pull/622)
> 
> That's strange, where does that Trailer without headers get created?

I had to dig to find it, but it seems to be created by some fallback logic I removed in [this](https://github.com/apache/pekko-grpc/pull/391/commits/82b9e57a8d9bdb9fb7d1f852c0a7cd0572986b87) commit

NavidJalali:
Looks like I missed something, I'll get back to debugging

interop-tests/src/test/scala/org/apache/pekko/grpc/interop/PekkoHttpServerProviderScala.scala
```diff
@@ -118,11 +118,10 @@ object PekkoHttpServerProviderScala extends PekkoHttpServerProvider with Directi
...
-          val origTrailers = response
+          response
             .attribute(AttributeKeys.trailer)
-            .map(_.headers)
-            .getOrElse(Vector.empty)
-            .map(e => RawHeader(e._1, e._2))
-          response.addAttribute(AttributeKeys.trailer, Trailer(f(origTrailers)))
+            .map(trailer => f(trailer.headers.map((RawHeader.apply _).tupled)))
+            .flatMap(headers => if (headers.isEmpty) None else Some(Trailer(headers)))
+            .fold(response)(response.addAttribute(AttributeKeys.trailer, _))
```
raboof, line 125: Ouch, good find!

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcExceptionDefaultHandleSpec.scala
```diff
@@ -58,7 +58,7 @@ class GrpcExceptionDefaultHandleSpec
             case Seq(LastChunk("", List(`Status`("3")))) => // ok
           }
         case _: Strict =>
-          response.attribute(AttributeKeys.trailer).get.headers.contains("grpc-status" -> "3")
+          response.headers.find(_.is("grpc-status")).map(_.value()) shouldBe Some("3")
```
raboof, line 61: The spec at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md says "Status must be sent in Trailers even if the status code is OK.". Perhaps we should allow both cases in this test? I.e. see if there is an `AttributeKeys.trailer`, if there is an `AttributeKeys.trailer` check that it has a grpc-status of 3 and the `response.headers` does *not* have a status, if there is no `AttributeKeys.trailer` then check the status in `response.headers`?

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcExceptionDefaultHandleSpec.scala
```diff
@@ -131,14 +131,18 @@ class GrpcExceptionDefaultHandleSpec
 
       val reply = GreeterServiceHandler(ExampleImpl).apply(request).futureValue
 
-      val lastChunk = reply.entity.asInstanceOf[Chunked].chunks.runWith(Sink.last).futureValue.asInstanceOf[LastChunk]
+      reply.entity shouldBe a[Strict]
+      val strict = reply.entity.asInstanceOf[Strict]
+      strict.contentType.mediaType.toString shouldBe "application/grpc+proto"
+      strict.data.isEmpty shouldBe true
+
```
raboof, line 138: similarly here, could we allow both the Strict and the Chunked case?

runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoNettyGrpcClientGraphStage.scala
```diff
@@ -112,6 +112,7 @@ private final class PekkoNettyGrpcClientGraphStage[I, O](
         override def onMessage(message: O): Unit =
           callback.invoke(message)
         override def onClose(status: Status, trailers: Metadata): Unit = {
+          if (!matVal.isCompleted) onHeaders(trailers)
           trailerPromise.success(trailers)
```
raboof, line 116: Hmm, this is a bit weird, putting the trailer headers *both* in the header *and* in the trailer field of the materialized value. I think we should probably do either - i.e. add an 'else'. 

That makes things a little more complicated for clients, who might have to look both at the headers and at the trailer headers to find e.g. the status. Perhaps we could create a helper function to make that easier at some point.

raboof (commented):
I like the general direction this is going, a couple of review points!

runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoNettyGrpcClientGraphStage.scala
NavidJalali: Hmm, yeah I thought about that too. If we do that, sometimes `trailerPromise` is never completed. I am not super familiar with all the things that are happening here but I was afraid of leaking that. If we do that should we not also do something with `trailerPromise`? Either fail it or succeed it with empty metadata?

NavidJalali: I thought at least this way, I am being more honest because the trailers and the headers are the same thing

raboof: `trailerPromise` never completing would indeed be bad, that's a great point!

Taking into account the spec at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md, the case where everything is in the first HEADERS segments is called the 'Trailers-Only' case - so perhaps it would be most faithful complete `onHeaders` with empty metadata, and complete the `trailerPromise` with the received headers?

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcExceptionDefaultHandleSpec.scala
NavidJalali: Trying to understand what you mean, it is allowed under the spec but knowing the implementation of the exception handler I know it will be a Strict. So adding another match arm that is never hit won't do much. But we _**should**_ probably make sure that such a test exist that HTTP Responses produced should 100% have trailers in one of the two possible locaitons. I don't know if this is the test to do it in tho, wdyt?

raboof: IMHO, a test should not describe "what the current implementation does", but "what the requirements on a correct implementation are": if you change the implementation from one 'correct' approach to another, ideally you shouldn't have to change the test. That said, if it makes the test too complex I'm OK with leaving it like this.

NavidJalali: That is a valid point, I agree. I will add the match arm here

runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoNettyGrpcClientGraphStage.scala
NavidJalali: Addressed in https://github.com/apache/pekko-grpc/pull/391/commits/b791d66abab4778d7cb178d4dc38e4dee3cf86e6

raboof (approved):
This looks good!

(the link validator failure is unrelated, #395)

NavidJalali:
@pjfanning is the link validator flakey? would it be green if you ran it again? I would love to get a snapshot build

pjfanning:
@NavidJalali don't worry about the link validator - we tend to ignore the results of that build - see https://github.com/apache/pekko-grpc/issues/395

pjfanning (approved):
lgtm

PR #396, opened 2024-10-30
Apply trailer only optimization for responseForSingleElement

NavidJalali:
I missed a case in https://github.com/apache/pekko-grpc/pull/391


raboof (approved):
Nice!

PR #401, opened 2024-11-03
set jvmopts

pjfanning:
* build takes a lot of memory - I'm having major issues with GCs when building locally
* this is a copy of the .jvmopts-ci


raboof (approved):

pjfanning:
causes CI build issues in our very complicated build

PR #402, opened 2024-11-03
grpc 1.68.1

pjfanning:
* try to workaround issues in #397 by not updating the grpc name-resolver listener in a Scala future
* there are some issues even after this - similar issues with grpc 1.68.1 enforcing that SynchronizationContext is used and that fails because we use Scala Futures
* https://github.com/grpc/grpc-java/issues/11662 I suspect the changes in grpc 1.68.1 are too big for us to workaround and that we might need to get grpc team to allow config to disable some of their recent changes


He-Pin:
the fix is included in 1.68.2

pjfanning:
replaced by #415

Issue #404, opened 2024-11-03
plugin-tester-scala does not use specified version of protoc


raboof:
It seems the `plugin-tester-scala` subproject inherits the protoc version from the sbt-protoc defaults instead of getting the one we intend to be using.

This can be fixed by adding a

```
PB.protocVersion := Dependencies.Versions.googleProtoc
```

to the plugin-tester-scala project, but I wonder if this shouldn't come in via `ReflectiveCodeGen` instead.

Not super urgent since it's a test project anyway, but might be worth having a look at at some point.

PR #400, opened 2024-11-03
link validator: ignore license report

raboof:
partly covers #395

PR #409, opened 2024-11-10
plugin-tester-scala: set protoc version

pjfanning:
see #404 - plugin-tester-java and interop-tests set this - so best to be consistent


raboof (approved):
(link validator failure is #400)

Issue #411, opened 2024-11-13
review need for pekko-grpc-codegen and bat files (published to maven central)


pjfanning:
https://repository.apache.org/content/groups/staging/org/apache/pekko/pekko-grpc-codegen_2.13/1.1.1/

See pekko-grpc-codegen_2.13-1.1.1-assembly.jar and pekko-grpc-codegen_2.13-1.1.1-bat.bat

* They are big files.
* I am not sure if they are needed
* The assembly jar bundles lots of scala-library, io.grpc, guava, protobuf, checkerframework classes but the only LICENSE and NOTICE in the jar are copied from the scala-library.
    * If we choose to keep publishing this file, we need to fix its build so that we add a proper LICENSE setup that properly attributes the license details for all classes bundled in the jar
* I don't know what the bat file is but presume it also includes compiled code from many sources

raboof:
> I am not sure if they are needed

They are used in https://github.com/apache/pekko-grpc/blob/main/gradle-plugin/src/main/groovy/org/apache/pekko/grpc/gradle/PekkoGrpcPlugin.groovy#L108 . 

> The assembly jar bundles lots of scala-library, io.grpc, guava, protobuf, checkerframework classes but the only LICENSE and NOTICE in the jar are copied from the scala-library.
> * If we choose to keep publishing this file, we need to fix its build so that we add a proper LICENSE setup that properly attributes the license details for all classes bundled in the jar

Ouch, indeed.

> I don't know what the bat file is but presume it also includes compiled code from many sources

I think it's basically the same as the assembly but wrapped in a way that makes it easier to execute on Windows

pjfanning:
> > I am not sure if they are needed
> 
> They are used in https://github.com/apache/pekko-grpc/blob/main/gradle-plugin/src/main/groovy/org/apache/pekko/grpc/gradle/PekkoGrpcPlugin.groovy#L108 .
> 
> > The assembly jar bundles lots of scala-library, io.grpc, guava, protobuf, checkerframework classes but the only LICENSE and NOTICE in the jar are copied from the scala-library.
> > 

Thanks @raboof for the pointer.
The jar plain pekko-grpc-codegen jar is published with a pom file that lists the dependencies.
https://repository.apache.org/content/groups/staging/org/apache/pekko/pekko-grpc-codegen_2.13/1.1.1/pekko-grpc-codegen_2.13-1.1.1.pom

[...]

PR #413, opened 2024-11-13
[WIP] Assembly jar LICENSE/NOTICE

pjfanning:
see #411 

Work in Progress. I still need to update the details for some dependencies.

Unfortunately, sbt-assembly does not appear to support adding extra files so the simplest seems to be change the LICENSE/NOTICE in the module jar and those will also appear in the assembly jars.


raboof (approved):
Cool! It would be nice to automate this more, but I can see how that'd be challenging. Being explicit like this is a good alternative.

Link validator failure is #400

PR #414, opened 2024-11-28
chore: fuse handlers in CancellationBarrierGraphStage

He-Pin:
Motivation:
Fuse the inHandler and outHandler to reduce memory usage.


raboof (approved):
I'm surprised this'd make a noticable difference, but the new code is at least as readable as the old, so :+1: 

I wonder if we should propose this component for inclusion in pekko-streams.

He-Pin:
Yeah, I was working on our internal grpc based implementation and just found this, it's more readable now.

PR #422, opened 2025-01-02
CI: add source headers

pjfanning:
header depends on whether the initial work is from Pekko times or if the yml is based on work that we inherited from Akka


He-Pin (approved):
lgtm

PR #447, opened 2025-04-18
Test with pekko snapshot

raboof:
To see if there's any surprising differences due to https://github.com/apache/pekko/pull/1795

not for merge


pjfanning:
With the gradle build failure due to mismatched pekko versions - possibly the quickest workaround is try to disable that version check.

pekko.fail-mixed-versions=off

raboof:
tbh the remaining tests succeeding makes me optimistic the change didn't have any negative effects, I don't think we need to test gradle specifically.

PR #456, opened 2025-05-27
preannounce trailers in server streaming calls

NavidJalali:
According to [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230) in section 4.4 regarding trailers it says:

> [4.4](https://www.rfc-editor.org/rfc/rfc7230#section-4.4).  Trailer
> When a message includes a message body encoded with the chunked transfer coding and the sender desires to send metadata in the form of trailer fields at the end of the message, the sender SHOULD generate a Trailer header field before the message body to indicate which fields will be present in the trailers. This allows the recipient to prepare for receipt of that metadata before it starts processing the body, which is useful if the message is being streamed and the recipient wishes to confirm an integrity check on the fly.
> Trailer = 1#field-name

When this pre-announcement some reverse proxy applications like tyk misbehave. So even if this is a **_SHOULD_** and not a **_MUST_**, I would appreciate if we could do this.

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
@@ -111,3 +111,27 @@ object `Status-Message` extends ModeledCustomHeaderCompanion[`Status-Message`] {
   def findIn(headers: immutable.Seq[HttpHeader]): Option[String] =
     headers.collectFirst { case h if h.is(name) => h.value() }
 }
+
+final class `Trailer` private (values: Seq[String]) extends ModeledCustomHeader[`Trailer`] {
+
+  override def companion: ModeledCustomHeaderCompanion[`Trailer`] = `Trailer`
+
+  override def value(): String = values.mkString(", ")
+
+  override def renderInRequests(): Boolean = true
+
+  override def renderInResponses(): Boolean = true
+}
+
+object `Trailer` extends ModeledCustomHeaderCompanion[`Trailer`] {
+  def apply(values: Seq[String]): `Trailer` = new `Trailer`(values.map(_.trim))
+
+  override val name = "Trailer"
+
+  override val lowercaseName: String = super.lowercaseName
+
+  override def parse(value: String): Try[`Trailer`] = Try(`Trailer`(value.split(',')))
+
+  def findIn(headers: immutable.Seq[HttpHeader]): Option[Seq[String]] =
+    headers.collectFirst { case header if header.is(name) => header.value().split(',').map(_.trim).toSeq }
+}
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcResponseHelpers.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcResponseHelpers.scala
@@ -97,7 +97,10 @@ object GrpcResponseHelpers {
 
   private def response(entity: Source[ChunkStreamPart, NotUsed])(implicit writer: GrpcProtocolWriter) = {
     HttpResponse(
-      headers = immutable.Seq(headers.`Message-Encoding`(writer.messageEncoding.name)),
+      headers = immutable.Seq(
+        headers.`Message-Encoding`(writer.messageEncoding.name),
+        headers.`Trailer`(headers.`Status`.name)
+      ),
       entity = HttpEntity.Chunked(writer.contentType, entity))
   }
 


runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
```diff
@@ -111,3 +112,29 @@ object `Status-Message` extends ModeledCustomHeaderCompanion[`Status-Message`] {
   def findIn(headers: immutable.Seq[HttpHeader]): Option[String] =
     headers.collectFirst { case h if h.is(name) => h.value() }
 }
+
+final class `Trailer` private (values: immutable.Seq[String]) extends ModeledCustomHeader[`Trailer`] {
```
raboof, line 116: OK to have it here for now, but perhaps also add a Pekko HTTP PR to model this header there?

pjfanning:
* this code does not compile in Scala 2.13 or 3.3
```
[error] -- Error: /home/runner/work/pekko-grpc/pekko-grpc/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala:133:68 
[error] 133 |  override def parse(value: String): Try[`Trailer`] = Try(`Trailer`(value.split(',')))
[error]     |                                                                    ^
[error]     |method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated since 2.13.0: implicit conversions from Array to immutable.IndexedSeq are implemented by copying; use `toIndexedSeq` explicitly if you want to copy, or use the more efficient non-copying ArraySeq.unsafeWrapArray
[error] one error found
```
* would it be possible to add tests?

runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
NavidJalali: I've made it private to scaladsl

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -182,6 +182,37 @@ abstract class PowerApiSpec(backend: String)
 
       metadataServer.terminate(3.seconds)
     }
+
+    "pre-announce trailers in the headers" in {
```
raboof, line 186: Please include a comment here as well explaining why it is important to pre-announce trailers: link to the RFC (though with the caveat that it's strictly speaking about chunked and not about h2) and mentioning the tyk bug. Maybe add an issue to https://github.com/TykTechnologies/tyk so you can link to it?

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -182,6 +182,37 @@ abstract class PowerApiSpec(backend: String)
...
+      val binding = Http()
```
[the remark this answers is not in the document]
NavidJalali: yep makes sense

runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
NavidJalali: I can take a look at opening a PR there this weekend

interop-tests/src/test/scala/org/apache/pekko/grpc/scaladsl/PowerApiSpec.scala
NavidJalali: Yeah I will need some help from my SRE colleagues to open an issue in there. They are on vacation now, but it would be good to come up with a fix for it as well

NavidJalali: sure I can move it 👀

NavidJalali:
Uhm, I am not sure what to put on the header of this new file. 
```
/*
 * Licensed to the Apache Software Foundation (ASF) under one or more
 * license agreements; and to You under the Apache License, version 2.0:
 *
 *   https://www.apache.org/licenses/LICENSE-2.0
 *
 * This file is part of the Apache Pekko project, which was derived from Akka.
 */
```
I guess since its new it doesnt need the lightbend part?

NavidJalali:
I guess technically lightbend doesnt even exist anymore its now called Akka 😂

pjfanning (approved):

pjfanning:
would this be a 1.2.0 or 1.1.2 change?

raboof:
> would this be a 1.2.0 or 1.1.2 change?

It's a feature, not a bugfix, so that'd make it 1.2.0 I think, right?

PR #457, opened 2025-05-27
test with scala 3.3.6

pjfanning:
* sbt setup was changed
* but we have GitHub hardcoding of version too

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -80,7 +80,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


raboof:
> > If we just make it `3.3` it should pick up the right patch version from sbt, so that might be neater still?
> 
> I think this is there for a reason, as in the build breaks when we don't have the full version. I'm 90%+ sure that I have attempted to fix this before.

Aah, it's needed for the codegen `publish`, and then perhaps it's fine to have it explicit elsewhere as well. Filed #458 to some day fix that ;)

raboof (approved):

PR #459, opened 2025-05-28
protect 1.1.x branch and add related CI jobs

pjfanning:
we have a 1.2 change ready for merge

### DIFF
--- a/.github/workflows/publish-1.1-nightly.yml
+++ b/.github/workflows/publish-1.1-nightly.yml
@@ -6,16 +6,16 @@
 # This file is part of the Apache Pekko project, which was derived from Akka.
 #
 
-name: Nightly Snapshot Dependency Test
+name: Publish 1.1 Nightly
 
 on:
   schedule:
-    - cron: "0 2 * * *"
+    - cron: "0 4 * * *"
   workflow_dispatch:
 
 jobs:
   sbt:
-    name: sbt test with latest pekko/pekko-main snapshots
+    name: sbt publish
     runs-on: ubuntu-latest
     if: github.repository == 'apache/pekko-grpc'
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-1.1-nightly.yml
```diff
@@ -0,0 +1,52 @@
...
+#   https://www.apache.org/licenses/LICENSE-2.0
+#
+# This file is part of the Apache Pekko project, which was derived from Akka.
+#
+
+name: Publish 1.1 Nightly
+
+on:
+  schedule:
+    - cron: "0 4 * * *"
```
raboof, line 13: isn't nightly a bit overkill?

raboof (approved):
I wonder if it wouldn't be sufficient to do a tag, and postpone creating a branch with protection and CI until we have a reason to make another 1.1.x release.

pjfanning:
> I wonder if it wouldn't be sufficient to do a tag, and postpone creating a branch with protection and CI until we have a reason to make another 1.1.x release.

* the next step after merging this is to tag the commit after the branch point as 1.2.0-M0 so that sbt-dynver produces versions for our snapshots that have 1.2.0 in their names - to differentiate from the 1.1 prefixed snapshot names currently generated
* I think it is easier to just cut the 1.1.x branch now instead of creating off of an old commit in the future
* this is consistent with what has been done before in this and other pekko repos
* I think we could as easily just aim to have the next release be a 1.2.0 instead of planning to do a 1.1.2 release that noone is asking for
* I can reduce the frequency of the publishes of snapshots for the legacy branches since we have workflow-dispatch enabled

raboof:
> I think we could as easily just aim to have the next release be a 1.2.0 instead of planning to do a 1.1.2 release that noone is asking for

Yes, I agree completely - which is why I wondered whether it's likely we'll end up needing the 1.1.x branch at all. But if you think it's worth it I'm fine with it :)

PR #471, opened 2025-06-24
add trailer attribute to trailer only optimized response

NavidJalali:
Trailer attribute is set by `GrpcProtocolWriter` in the success path but in the failure path it is missing

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcResponseHelpers.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcResponseHelpers.scala
@@ -106,11 +106,11 @@ object GrpcResponseHelpers {
   }
 
   def status(trailer: Trailers)(implicit writer: GrpcProtocolWriter): HttpResponse = {
+    val trailerHeaders = GrpcEntityHelpers.trailers(trailer.status, trailer.metadata)
     HttpResponse(
       headers =
-        headers.`Message-Encoding`(writer.messageEncoding.name) ::
-        GrpcEntityHelpers.trailers(trailer.status, trailer.metadata),
+        headers.`Message-Encoding`(writer.messageEncoding.name) :: trailerHeaders,
       entity = HttpEntity.empty(writer.contentType)
-    ).withAttributes(Map(AttributeKeys.trailer -> trailer))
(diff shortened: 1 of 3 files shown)


NavidJalali:
Ah it seems I made a bit mistake 😄 
There is a difference in types

runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -111,6 +111,6 @@ object GrpcResponseHelpers {
         headers.`Message-Encoding`(writer.messageEncoding.name) ::
         GrpcEntityHelpers.trailers(trailer.status, trailer.metadata),
       entity = HttpEntity.empty(writer.contentType)
-    )
+    ).withAttributes(Map(AttributeKeys.trailer -> trailer))
```
raboof, line 114: if the trailer headers are already written to the `headers`, are they still needed as `attribute`?

NavidJalali: I think its nicer on the akka http level to have a unified way of finding the trailer regardless of the success or failure of the request, wdyt

raboof: AFAICT this change would cause the status/metadata headers to be added twice: once to the regular response, and then once again as a block of trailing headers.

While I can understand the attraction of having the status information in the 'same place', I'm not sure that weights up against the optimization of having a regular set of headers rather than a split between normal and trailing headers in this case.

NavidJalali: Wait is that because I manually add them to the headers myself and if a trailer attribute gets detected it will also get added somewhere further in the pipeline?

NavidJalali: Wouldn't it be preferable to then only add it to the attributes and rely on it getting added later on?

NavidJalali: Can you point me to the place where this translation from trailer attributes to headers happens?

Issue #483, opened 2025-08-04
remove of pekko.japi.Function


pjfanning:
Replaced by pekko.japi.function.Function
https://github.com/search?q=repo%3Aapache%2Fpekko-grpc+pekko.japi&type=code

PR #484, opened 2025-08-04
try to avoid deprecated japi function

pjfanning:
see #483

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/javadsl/package.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/javadsl/package.scala
@@ -19,7 +19,7 @@ package object javadsl {
 
   /**
    * Helper for creating org.apache.pekko.japi.function.Function instances from Scala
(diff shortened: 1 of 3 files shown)


runtime/src/main/scala/org/apache/pekko/grpc/javadsl/package.scala
```diff
@@ -40,20 +40,20 @@ package object javadsl {
     }
 
   /**
-   * Helper for creating Scala partial functions from [[pekko.japi.Function]]
-   * instances as Scala 2.11 does not know about SAMs.
+   * Helper for creating Scala partial functions from [[pekko.japi.function.Function]]
+   * instances.
    */
   @deprecated("no longer needed since support for Scala 2.11 has been dropped", "1.2.0")
```
He-Pin, line 46: Should we remove this too

pjfanning: I'll have a look at this again. I think it is currently needed.

PR #485, opened 2025-08-04
remove function wrapping and deprecate unused function wrapping functions

pjfanning:
* some of the wrapping may only have been needed in Scala 2.11
* tidy up related to #483 but this change is safe for a 1.3.0 release


He-Pin (approved):
lgtm

PR #493, opened 2025-09-01
chore: Update PekkoCoreDependency version to 1.2.0

He-Pin:
Test with 1.2.0


pjfanning:
See https://github.com/search?q=repo%3Aapache%2Fpekko-grpc+1.1.5&type=code

This build is more complicated than other builds. The Pekko version appears in multiple places.

He-Pin:
trait Function in package japi is deprecated (since 1.2.0): Will be removed in favor of pekko.japi.function.Function in Pekko 2.0.0

raboof:
main is actually already at 1.3.0 by now, so this seems outdated

PR #494, opened 2025-09-01
chore: Drop unused deprecated pekko.japi.Function

He-Pin:
Motivation:
pekko.japi.Function will be removed in Pekko 2.0.0


pjfanning:
Aim would be to publish a 2.0.0-M1 release for Pekko Core in the next few weeks. So we could start dev on pekko-grpc 2.0.0 then.

In meantime, for pekko-grpc 1.2, should we add annotations to suppress deprecation warnings?

pjfanning:
@He-Pin I created a 1.2.x branch so we can now start making breaking changes in main branch - do you want to add the mima filters to this PR so we can get it merged?

PR #495, opened 2025-09-04
use 1.2.x snapshots in nightly test workflow

pjfanning:
* Pekko 2 is matching 'main' and that has breaking build changes
* Safer to test with latest 1.2.x snapshots
* https://github.com/apache/pekko-grpc/actions/workflows/nightly-snapshot-dependency-test.yml


He-Pin (approved):
lgtm

PR #500, opened 2025-09-09
protect 1.2.x branch

pjfanning:
* we have changes set for a 1.2.0 release but it doesn't feel urgent
* we can make progress getting 2.0.0 oriented changes in to main branch if we have a dedicated 1.2.x branch


He-Pin (approved):
Lgtm

PR #502, opened 2025-09-10
remove deprecated methods

pjfanning:
targeted at 2.0.0

PR #503, opened 2025-09-10
improve performance of deserializing ByteString

pjfanning:
* also done on main branch as part of #502 
* ByteStringInputStream avoids cloning of the byte data in most cases


raboof (approved):
cool - but conflicting with main?

raboof:
> Aplogies, I meant to target this at branch 1.2.x. That is fixed now.

ah it was right there in the pr message :smile: :+1:

PR #506, opened 2025-09-11
make Scala 2.13.16 the default for the CI build and drop Scala 2.12 (except for sbt plugins)

pjfanning:
* Pekko 2.0.0 drops Scala 2.12 support
* We still need to publish Scala 2.12 jars for sbt plugins - for sbt 1.x


raboof (approved):
Nice!

I think we wait with merging this until pekko core 2.x is out? or do we already want to start breaking things on main? I'm fine with either.

PR #565, opened 2025-12-11
upgrade to latest scala (2.13.18, 2.12.21, 3.3.7) (#560)

pjfanning:
need to fix issues in https://github.com/apache/pekko-grpc/actions/workflows/nightly-snapshot-dependency-test.yml

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -80,7 +80,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        SCALA_VERSION: [2.12.20, 2.13.16, 3.3.6]
+        SCALA_VERSION: [2.12.21, 2.13.18, 3.3.7]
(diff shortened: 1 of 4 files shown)


sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/build.sbt
```diff
@@ -7,9 +7,7 @@
  * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
-// Can be removed when we move to 2.12.14
-// https://github.com/akka/akka-grpc/pull/1279
-scalaVersion := "2.12.20"
+scalaVersion := "2.12.21"
```
raboof: are we sure we still need the explicit scala version here? the comment/issue seems to suggest it was only added because of a specific problem with 2.12.13

PR #566, opened 2025-12-11
run sbt test that uses scala3

pjfanning:
I can't see any evidence that this test is run. Move it in with the other scala sbt tests so it will run with them.


pjfanning:
#567 worked better

PR #568, opened 2025-12-12
Pekko HTTP snapshot

raboof:
trying a variation on #563


raboof:
(we do this usually, but especially for this one please squash when merging :) )

PR #592, opened 2026-01-14
Add 'using' for implicit args in Scala 3 sources.

timw:
Generate 'using' keyword for explicitly invoked implicit arg lists.
Silences warnings in Scala 3 and -Xsource:3 compilation.


pjfanning:
Thanks for the PR. Is there a way that an automated test can be added or something that would break if the generated code is incorrect?

pjfanning:
This is an enhancement to #222

mdedetrich:
> Thanks for the PR. Is there a way that an automated test can be added or something that would break if the generated code is incorrect?

If the code would be incorrect then for all intents and purposes the compiler would be broken. The best way would be to check generate JVM bytecode before and after but I would be shocked if there would be a difference as its just a syntax alias.

PR #602, opened 2026-01-21
remove use of differentiated scala 2.12/2.13 code

pjfanning:
the only diff in the code was the `nowarn` needed by 2.12 but I see no harm in this being used in a common source file


raboof (approved):
I have no strong preference

PR #615, opened 2026-01-29
gRPC directives for pekko-http

timw:
This adds pekko-http directives that detect grpc/grpc-web requests.

This support integration of grpc endpoints alongside other HTTP requests on the same server binding (e.g. HTTP based /healthz handlers etc.) and different handling for different protocol types (e.g. enforcing `Strict-Transport-Security` for grpc-web requests).

Java and Scala DSLs are provided.
Currently the Java DSL uses the `@InternalApi` `RouteHandler` from pekk-http - this could be copied in-tree if that's not considered maintainable.

The `GrpcProtocol.detect` hasn't been used, as it doesn't check the main type of the MediaType at present - alternatively that could be changed and used for this purpose as well.


runtime/src/test/scala/org/apache/pekko/grpc/scaladsl/GrpcDirectivesSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.grpc.scaladsl
+
+import org.apache.pekko
```
pjfanning: Can you move keep the pekko imports together?

timw:
Roughly the guts of our server route looks like this:

```
concat(

  grpc {
    nocache {
      grpcSpan(otelTracer) {
          grpcServicesRoute
      }
    }
  },

  grpcWeb {
    nocache {
      handleRejections(corsRejectionHandler) {
        cors(corsSettings) {
          requireSTS {
            grpcSpan(otelTracer) {
              grpcServicesRoute
            }
          }
        }
      }
    }
  },

  httpRoutes // <-- includes HTTP tracing around specific non-gRPC HTTP services
)
```

Ultimately this is just an ergonomic mapping on media types, so I can just internalise in my app, but as you can see you can quite quickly run into cases where particular directives should/must be applied to particular media type branches.

For some it's correctness/safety (i.e. I don't want to apply STS headers and CORS for `grpc` requests), but some directives require up-front work and have a penalty if an inner route rejects a request (e.g. there's no API to cancel an OTEL span once create/registered, so you have to mark it as discarded and filter it in the exporter to suppress it).

PR #653, opened 2026-03-08
chore: migration trace - send transitive dependencies in server reflection

He-Pin:
## Motivation

This PR tracks the migration of upstream akka/akka-grpc commit `bf1e13c8` which adds transitive dependency tracking to the gRPC Server Reflection implementation.

The gRPC Server Reflection protocol requires that responses include not just the requested file descriptor but also all its transitive proto dependencies. Without this, clients using reflection (e.g., grpcurl, Postman gRPC) may fail to resolve types referenced in service definitions.

> **Note:** This feature was independently ported to `apache/pekko-grpc` as **PR #643** (merged as commit `7a954161`). This PR exists solely for migration audit trail purposes.

## Modification

- Refactored `ServerReflectionImpl.serverReflectionInfo` to use `statefulMap` with a `Set[String]` state to track already-sent descriptors per stream, preventing duplicate sends
- Added `withTransitiveDeps` helper to collect full transitive closure of proto dependencies
- Added `toFileDescriptorResponse` helper to convert descriptors to reflection responses
- Added 4 test proto files forming an import chain for testing
- Added 2 new test cases verifying transitive deps are returned exactly once

## Result

Server reflection correctly returns all transitive proto dependencies, enabling gRPC clients that rely on reflection to fully resolve complex proto type hierarchies.

## References

- Upstream commit (**which is now Apache licensed**): https://github.com/akka/akka-grpc/commit/bf1e13c8
- Upstream PR: https://github.com/akka/akka-grpc/pull/1719
- Already merged in pekko-grpc as: https://github.com/apache/pekko-grpc/pull/643
- Our local migration commit: `b5722812`


He-Pin:
## Local Investigation Results: Why This PR Has No File Changes

I've run a thorough local investigation to understand why this PR contains zero file changes.

**Finding**: The server reflection transitive dependencies feature (ported from akka/akka-grpc commit [`bf1e13c8`](https://github.com/akka/akka-grpc/commit/bf1e13c8), which is now Apache licensed) was **independently implemented and merged** into apache/pekko-grpc as [PR #643](https://github.com/apache/pekko-grpc/pull/643) (commit `7a954161`).

**Verification** — comparing our migration commit `b5722812` against `upstream/main` shows **zero diff** in all affected files:
- `runtime/src/main/scala/org/apache/pekko/grpc/internal/ServerReflectionImpl.scala` — identical
- `runtime/src/test/protobuf/pekko/grpc/internal/reflection_test_*.proto` — identical  
- `runtime/src/test/scala/org/apache/pekko/grpc/internal/ServerReflectionImplSpec.scala` — identical

Since the feature is already present in `main` with identical code, there is no incremental value in merging this PR. Closing as the migration goal for this commit is already achieved via PR #643.

Issue #698, opened 2026-05-17
Typed Actors?


kazuf3:
I'm walking through the documents these days and have been struggled for a while to make my scenario work with Pekko gRPC.
The main difficulty for me came from the difference between typed and classic actors, because I'm a Akka/Pekko newbie.
Especially the difference between the walkthrough document and g8 quickstart code examples.

Walkthrough:
- uses classic actor
- lacks of test code

g8 quickstart code:
- uses typed actor
- comes with test code

Now I think I would like to offer migrating the walkthrough code to typed actors.
However, after a brief research, I found the dependency difference between this repository and g8 sample.
g8 sample has dependency to:
`  "org.apache.pekko" %% "pekko-actor-typed" % pekkoVersion,`
while in this repository we don't have any dependency to it.

If I change the code in plugin-tester-scala to typed, probably I will need the typed actor library dependency somewhere in project/*Dependency.scala.

Another option is to add test code for classic actor, something like `src/test/scala/example/myapp/statefulhelloworld/GreeterActorSpec.scala`.

Any thoughts?

pjfanning:
#702 merged

PR #699, opened 2026-05-17
migrated to typed actor in plugin-tester-scala; it will be also used …

kazuf3:
…for doc: server/walkthrough/stateful services section

(See #698)


kazuf3:
Do I also have to update build.gradle?

pjfanning (commented):
can't we keep the legacy actor and create a new sample using the typed actor?

pjfanning:
> can't we keep the legacy actor and create a new sample using the typed actor?

I would suggest leaving https://github.com/apache/pekko-grpc/tree/main/plugin-tester-scala/src/main/scala/example/myapp/helloworld as is and adding a new typedhelloworld package in the same plugin-tester-scala module that uses the typed actor.

kazuf3:
OK, ill create a new branch/PR with:
- new class path/file names
- change the document reference annotation

kazuf3:
Updated PR to #702. Closing this.

Issue #677, opened 2026-04-08
Add support for sbt 2


mkurz:
`pekko-grpc-sbt-plugin` should be published for sbt 2 as well:
- https://repo1.maven.org/maven2/org/apache/pekko/pekko-grpc-sbt-plugin_2.12_1.0/

/cc @xuwei-k

pjfanning:
* https://www.scala-lang.org/blog/2026/03/02/sbt2-compat.html
* https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html

pjfanning:
I don't think we can make much progress until scalapb sbt plugins are published for Sbt 2 / Scala 3

https://repo1.maven.org/maven2/com/thesamet/ (especially sbt-protoc) -- https://github.com/thesamet/sbt-protoc/issues/422

mkurz:
Needed to make play 2 samples pass sbt 2:
- https://github.com/playframework/play-samples/pull/1027

xuwei-k:
https://repo1.maven.org/maven2/com/thesamet/sbt-protoc_sbt2_3/

He-Pin:
That's great. Let's wait until Scala moves to sbt2?

raboof:
> That's great. Let's wait until Scala moves

Ideally we could support both

mkurz:
> > That's great. Let's wait until Scala moves to sbt2?
> 
> Ideally we could support both

What do you mean? The `pekko-grpc-sbt-plugin` sbt plugin? Should be totally doable to support both sbt1 and sbt2 - lots of sbt plugins already support both.

pjfanning:
https://github.com/apache/pekko-grpc/pull/610 is the next issue that needs to be resolved but ideally scalapb will release a 1.0.0 version so that we don't have to rely on the alpha4 version.

mkurz:
Great! Can you tag? Thanks!

pjfanning:
It will be a few weeks before do a release with this and they will likely to be a milestone release for 2.0.0.

Snapshots can be found at
https://repository.apache.org/content/repositories/snapshots/org/apache/pekko/pekko-grpc-sbt-plugin_sbt2_3/

The version number will match all the other pekko-grpc libs.

[...]

PR #713, opened 2026-05-29
Codegen fix

pjfanning:
* CI build 'nightly' is failing since #677 
* this is a band aid but I'm hoping it will allow us to get some testing of the sbt 2 plugin support

### DIFF
--- a/codegen/src/main/scala/org/apache/pekko/grpc/gen/Main.scala
+++ b/codegen/src/main/scala/org/apache/pekko/grpc/gen/Main.scala
@@ -16,12 +16,15 @@ package org.apache.pekko.grpc.gen
 import java.io.ByteArrayOutputStream
 import java.net.URLDecoder
 
+import scala.annotation.nowarn
+
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import org.apache.pekko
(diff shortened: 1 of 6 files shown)


codegen/src/main/scala/org/apache/pekko/grpc/gen/Main.scala
```diff
@@ -16,12 +16,15 @@ package org.apache.pekko.grpc.gen
...
 
+import scala.annotation.nowarn
+
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import org.apache.pekko
 import pekko.grpc.gen.javadsl.{ JavaClientCodeGenerator, JavaInterfaceCodeGenerator, JavaServerCodeGenerator }
 import pekko.grpc.gen.scaladsl.{ ScalaClientCodeGenerator, ScalaServerCodeGenerator, ScalaTraitCodeGenerator }
 
 // This is the protoc plugin that the gradle plugin uses
+@nowarn("msg=deprecated")
```
pjfanning, line 27: this code needs to compile with Scala 3.8 now (sbt 2 plugins needs Scala 3.8+)
but also needs to compile in Scala 2.12
code compiles in Scala 3.8 but 'extends App' is now deprecated

He-Pin (approved):

PR #732, opened 2026-06-18
fix: remove -Yfuture-lazy-vals for Scala 3.9+

He-Pin:
### Motivation
Scala 3.9.0-RC1 removed the `-Yfuture-lazy-vals` compiler option (the behavior is now the default). Under `-Werror`, the "bad option" warning becomes a compilation error.

### Modification
Conditionally include `-Yfuture-lazy-vals` only for Scala 3.x where minor version < 9, using `CrossVersion.partialVersion` check.

### Result
pekko-grpc codegen module compiles cleanly with Scala 3.9.0-RC1.

### Tests
- `sbt "++3.9.0-RC1!; codegen/compile"` passes

### References
None - Scala 3.9 forward compatibility

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -54,14 +54,8 @@ object Common extends AutoPlugin {
            "-encoding",
            "UTF-8")
        else
-         Seq(
-           "-unchecked",
-           "-deprecation",
-           "-Werror",
-           "-Wunused:imports",
-           "-Yfuture-lazy-vals",
-           "-encoding",
-           "UTF-8")),
+         Seq("-unchecked", "-deprecation", "-Werror", "-Wunused:imports", "-encoding", "UTF-8") ++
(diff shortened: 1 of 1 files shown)


project/Common.scala
```diff
@@ -54,14 +54,9 @@ object Common extends AutoPlugin {
...
-         Seq(
-           "-unchecked",
-           "-deprecation",
-           "-Werror",
-           "-Wunused:imports",
-           "-Yfuture-lazy-vals",
-           "-encoding",
-           "UTF-8")),
+         Seq("-unchecked", "-deprecation", "-Werror", "-Wunused:imports", "-encoding", "UTF-8") ++
+         (if (CrossVersion.partialVersion(scalaVersion.value).exists(_._2 < 9)) Seq("-Yfuture-lazy-vals")
```
mkurz, line 58: what if Scala 3.9.10 is used? that also matches `< 9` for minor version

mkurz: ah sorry i am wrong, forget it

pjfanning (approved):
lgtm

PR #736, opened 2026-06-19
sbt-plugin: publish in non-legacy style

raboof:
We enabled the legacy style this in the past (#543) to keep support for sbt 1.8 and below.

Now that `main` targets a future 2.0.0 release, I think it's reasonable to move to the non-legacy style there. Notably the Maven Portal no longer supports the old style, and while _currently_ we're not affected by that as we deploy through `repository.a.o` this seems risky.


pjfanning (approved):
lgtm

PR #739, opened 2026-06-21
Optimize gRPC request processing: fast negotiation path and reduced per-request allocations

He-Pin:
## Motivation

Profiling the gRPC unary request hot path revealed several allocation and CPU overhead sources:

1. **Protocol negotiation**: `GrpcProtocol.negotiate()` performed redundant header scans via separate `detect()` + `Codecs.detect()` + `Codecs.negotiate()` calls per request
2. **Per-request allocations**: Lambda objects, `ByteString.slice` wrappers, `Option.map` closures, and `PartialFunction` allocations in the unary handler path
3. **URI path matching**: `methodName` used backtick pattern matching which could be slightly improved with guard-based matching

### Protocol negotiation (3 commits)
- **Fast negotiation path**: Single-pass header scan for native gRPC (`application/grpc+proto`), avoiding redundant `detect()` + `Codecs.detect()` + `Codecs.negotiate()` calls
- **Respect client compression preference**: Writer codec respects `grpc-accept-encoding` header, using the client's preferred codec order
- **String comparison fix**: Changed `eq` to `==` for proper string comparison in negotiate fast path
- **Pre-computed negotiation result**: Cache `NativeIdentityIdentity` tuple to avoid per-request allocation
- **`negotiated` lambda elimination**: Replace `Option.map` with direct `match` to avoid closure allocation

### Unary hot path (6 commits)
- **Pre-create unary method lambdas**: Generated handler pre-creates lambdas for unary methods at initialization, avoiding per-request lambda allocation
- **Native identity strict decoding**: Optimized `decodeIdentitySingleFrame` to parse gRPC frame header directly without `ByteReader` allocation
- **Offset-aware deserialization**: Added `ProtobufFrameSerializer.deserialize(data, offset, length)` to bypass `ByteString.slice` allocation (~2 wrapper objects per request)
- **`methodName` guard-based matching**: Replaced backtick pattern matching with guard condition for slightly cleaner generated code
- **Inline exception handling**: `handleUnaryException` inlines the error-to-response conversion, avoiding `PartialFunction` allocation via `GrpcExceptionHandler.from()`
- **`encodeFrameData` small message optimization**: For messages ≤4KB, combine frame header + data into single array allocation

## Result

Benchmarked with `ghz` (complex_proto, 1000 concurrency, 50 connections, 120s, SerialGC):

| Metric | Baseline (1.2.0) | Optimized | Improvement |
|--------|-----------------|-----------|-------------|
| Throughput (ForkJoinPool) | 62,899 req/s | 73,584 req/s | **+17.0%** |
| P99 latency (ForkJoinPool) | 32.46 ms | 29.20 ms | **-10.0%** |
| Avg latency (ForkJoinPool) | 8.34 ms | 6.77 ms | **-18.8%** |
| P99 latency (AffinityPool) | — | 24.17 ms | **-25.5%** |

JMH microbenchmarks confirm:
- `GrpcMarshallingBenchmark.unmarshallStrict`: ~50M ops/s
- `ScalaUnaryHandlerBenchmark.generatedUnary`: ~3.8M ops/s

## Tests

- `sbt runtime / Test / testOnly GrpcProtocolSpec GrpcMarshallingSpec CodecsSpec GrpcResponseHeadersSpec`
- `sbt codegen / Test / test`
- All tests pass

## References

- Profiling with async-profiler (CPU + allocation)
- Benchmark suite: [grpc_bench](https://github.com/LesnyRumcajs/grpc_bench) scala_pekko_bench


pjfanning:
@He-Pin can the gzip change be moved to its own PR because it is a behaviour change - better for transparency?

He-Pin:
I agree, I think we can add an adaptive one for it,wdyt @pjfanning

pjfanning:
@He-Pin if you want to make the compression adaptive then that's ok by me.

He-Pin:
Let's do that on this PR. at $work, it's adaptive.

He-Pin:
@pjfanning I have removed the gzip change from this PR

runtime/src/main/scala/org/apache/pekko/grpc/internal/Codecs.scala
```diff
@@ -23,8 +23,9 @@ import scala.collection.immutable
 import scala.util.{ Failure, Success, Try }
 
 object Codecs {
-  // TODO should this list be made user-extensible?
-  val supportedCodecs = immutable.Seq(Gzip, Identity)
+  // Codecs supported by this server. The selection order is determined by the
+  // client's grpc-accept-encoding header preference, not by this sequence order.
+  val supportedCodecs = immutable.Seq(Identity, Gzip)
```
pjfanning, line 28: can this gzip change be handled only in the new separate PR?

runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcProtocolNative.scala
```diff
@@ -44,11 +44,32 @@ object GrpcProtocolNative extends AbstractGrpcProtocol("grpc") {
     AbstractGrpcProtocol.writer(this, codec, encodeFrame(codec, _), encodeDataToResponse(codec))
 
   override protected def reader(codec: Codec): GrpcProtocolReader =
-    AbstractGrpcProtocol.reader(codec, decodeFrame)
+    if (codec eq Identity)
```
pjfanning, line 47: this identity codec change, can it be done in the gzip PR instead?

runtime/src/main/scala/org/apache/pekko/grpc/internal/Codecs.scala
He-Pin: Will rebate, I think I will continues it this weekends

PR #755, opened 2026-06-26
refactor: use Files.readString (Java 11+)

He-Pin:
### Motivation
`Files.readString` was introduced in Java 11 and provides a cleaner API for reading an entire file into a `String`, replacing the `new String(Files.readAllBytes(path), charset)` pattern.

### Modification
Replace `new String(Files.readAllBytes(path), "UTF-8")` with `Files.readString(path, StandardCharsets.UTF_8)` in:
- `benchmark-java/src/main/java/org/apache/pekko/grpc/benchmarks/Utils.java`
- `interop-tests/src/main/java/org/apache/pekko/grpc/interop/PekkoGrpcServerJava.java`

### Result
More idiomatic Java 11+ code with identical behavior.

### Tests
- [ ] Existing tests pass (behavior-preserving refactoring)
- [ ] `sbt javafmtAll` passes with JDK 17

### References
None - Java 11+ API migration

### DIFF
--- a/benchmark-java/src/main/java/org/apache/pekko/grpc/benchmarks/Utils.java
+++ b/benchmark-java/src/main/java/org/apache/pekko/grpc/benchmarks/Utils.java
@@ -37,6 +37,7 @@ import java.io.*;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.SocketAddress;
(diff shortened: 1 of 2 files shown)


pjfanning:
javafmt broken

pjfanning (approved):
lgtm

PR #756, opened 2026-06-26
refactor: use List.copyOf/Set.copyOf (Java 10+)

He-Pin:
## Summary
- Replace `Collections.unmodifiableList(new ArrayList<>(...))` with `List.copyOf(...)` in ServerConfiguration
- Replace `Collections.unmodifiableSet(new LinkedHashSet<>(...))` with `Set.copyOf(...)` in ClientConfiguration

## Changes
- `codegen/src/main/scala/org/apache/pekko/grpc/gen/javadsl/ServerConfiguration.java`
- `codegen/src/main/scala/org/apache/pekko/grpc/gen/javadsl/ClientConfiguration.java`

## Notes
- Source collections derive from enum `.values()` arrays, so null elements are impossible
- `List.copyOf`/`Set.copyOf` rejecting nulls is a beneficial fail-fast property for these config classes
- Removed unused `Collections` and `LinkedHashSet` imports from ClientConfiguration


pjfanning (approved):
lgtm

PR #770, opened 2026-07-02
Generate Scala 3-friendly gRPC sources

### LINKED ISSUE #766: Code generator produces Scala 3 unfriendly code (incomplete Scala 3 source generation)
### Motivation

The pekko-grpc code generator has **partial** Scala 3 support via `ScalaCompatConstants` and the `scala3_sources` flag, but the implementation is incomplete. Several Twirl templates still emit Scala 2-only syntax when `scala3_sources: true` is set, and the `scala3_sources` flag is not accessible from Maven or Gradle plugins.

#### 1. `Marshallers.scala.txt` has no Scala 3 awareness

`codegen/src/main/twirl/templates/ScalaCommon/Marshallers.scala.txt` never references `service.scalaCompatConstants`. All `implicit val` / `implicit def` declarations are emitted unconditionally regardless of the `scala3_sources` flag.

Affected definitions (lines 39, 42, 45, 48, 51):
```scala
implicit val @serializer.name: ScalapbProtobufSerializer[...] = ...
implicit def unmarshaller[T](implicit serializer: ProtobufSerializer[T], mat: Materializer) = ...
implicit def toSourceUnmarshaller[T](implicit serializer: ProtobufSerializer[T], mat: Materializer) = ...
implicit def marshaller[T](implicit serializer: ProtobufSerializer[T], writer: GrpcProtocolWriter, system: ClassicActorSystemProvider) = ...
implicit def fromSourceMarshaller[T](implicit serializer: ProtobufSerializer[T], writer: GrpcProtocolWriter, system: ClassicActorSystemProvider) = ...
```

#### 2. `implicit` not switched to `using` at definition sites

`ScalaCompatConstants.ImplicitUsing` is only applied at **call sites** in `ScalaServer/Handler.scala.txt` (lines 133, 135, 137, 176, 178, 180). The following **definition sites** still emit `implicit` unconditionally:

- **`ScalaClient/Client.scala.txt:56,58,61,64`** — `(implicit sys: ClassicActorSystemProvider)` and `private implicit val ex`
- **`ScalaServer/Handler.scala.txt:43,53,67,79,91,104,118,161`** — `(implicit system: ClassicActorSystemProvider)` on `apply` / `withServerReflection` / `handler` / `partial` methods
- **`ScalaServer/Handler.scala.txt:119,120,162,163`** — `implicit val mat` and `implicit val ec`
- **`ScalaCommon/Marshallers.scala.txt`** — all `implicit` definitions (see above)

In Scala 3, `implicit` still compiles but generates deprecation warnings. With `-Xfatal-warnings` (as used in `sbt-plugin/src/sbt-test/scala3/02-scala3-sourcegen/build.sbt:12`), these warnings become errors.

#### 3. `WildcardType` constant is commented out

`codegen/src/main/scala/org/apache/pekko/grpc/gen/scaladsl/ScalaCompatConstants.scala:21`:
```scala
// val WildcardType: String = if (emitScala3Sources) "?" else "_"
```

If any template ever needs to emit a wildcard type parameter (e.g., `Foo[_]` → `Foo[?]`), the switch is not available.

#### 4. Maven plugin does not support `scala3_sources`

`maven-plugin/src/main/scala/org/apache/pekko/grpc/maven/AbstractGenerateMojo.scala:92-97` — `parseGeneratorSettings` does not propagate a `scala3_sources` key. Maven users cannot enable Scala 3 source generation via plugin configuration.

#### 5. Gradle plugin does not support `scala3_sources`

`gradle-plugin/src/main/groovy/org/apache/pekko/grpc/gradle/PekkoGrpcPlugin.groovy:128-152` — the hardcoded options list (`generate_client`, `generate_server`, `server_power_apis`, `use_play_actions`, `extra_generators`, etc.) does not include `scala3_sources`. Gradle users can only activate it via the proto file option.

#### 6. Call-site `using` with definition-site `implicit` mismatch

When `emitScala3Sources=true`, the handler template generates `using` at call sites (e.g., `GrpcMarshalling.handleUnary(...)(using deserializer, serializer, mat, reader, writer, system, ec)`), but the runtime library methods (e.g., `GrpcMarshalling.handleUnary` at `runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/GrpcMarshalling.scala:114-124`) are defined with `implicit` parameter lists. While Scala 3 allows this, it is inconsistent and may cause confusion.

#### 7. ScalaPB-generated code contains `private[this]` and `implicit class`

While this is a ScalaPB issue rather than a pekko-grpc issue, pekko-grpc users targeting Scala 3 are affected. ScalaPB-generated message classes contain:
- `private[this]` (instance-private access, deprecated in Scala 3)
- `implicit class` for lens support (deprecated in Scala 3, should be `extension` methods)
- `@SerialVersionUID(0L)` on case classes (generates warnings)

This is tracked upstream in ScalaPB and may already be resolved in newer versions, but pekko-grpc should ensure it depends on a Scala 3-compatible ScalaPB version.

### Modification

The following changes would address these issues:

1. **`Marshallers.scala.txt`**: Add `service.scalaCompatConstants` references for `implicit` → `using` switching at definition sites, similar to how `Handler.scala.txt` does it for call sites.
2. **All templates**: Switch `implicit` to `using` at definition sites when `emitScala3Sources=true`.
3. **`ScalaCompatConstants`**: Uncomment and implement `WildcardType`.
4. **Maven plugin**: Add `scala3_sources` support to `AbstractGenerateMojo.parseGeneratorSettings`.
5. **Gradle plugin**: Add `scala3_sources` to the options list in `PekkoGrpcPlugin`.
6. **Runtime library**: Consider migrating `implicit` parameter lists to `using` (this is a larger effort that affects binary compatibility).

### Result

Generated code would be fully Scala 3 compatible when `scala3_sources: true` is set, without deprecation warnings or compilation errors under `-Xfatal-warnings`.

### References

- Related: #337 (Gradle plugin Scala 2.13 compatibility)
- ScalaPB Scala 3 support: https://scalapb.github.io/docs/scalapb-3
- Scala 3 migration guide: https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html

He-Pin:
### Motivation
Fixes #766. Generated Scala gRPC sources still emitted Scala 2 implicit syntax when `scala3_sources` was enabled, and Gradle users had no direct `scala3Sources` option.

### Modification
Generate Scala 3-friendly `using`/`given` syntax and wildcard imports from the Scala templates when `scala3_sources` is enabled. Add Gradle `scala3Sources` wiring for both `pekkoGrpc` and `scalapb`, cover Maven generator setting parsing, and document Gradle and Maven configuration. The Gradle unit test inspects protobuf plugin options through a `VarHandle` helper instead of reflective `Method#setAccessible` access.

### Result
Generated Scala sources can opt into Scala 3 syntax consistently across sbt, Gradle, and Maven configuration paths.

### Tests
- `sbt "sbt-plugin/scripted scala3/02-scala3-sourcegen"` / environment failed before publishing local Scala 3 runtime: missing `org.apache.pekko:pekko-grpc-runtime_3` SNAPSHOT
- `sbt "codegen / Compile / compile"` / passed
- `sbt "maven-plugin / Test / compile"` / passed
- `sbt "maven-plugin / Test / testOnly org.apache.pekko.grpc.maven.ProtocSpec"` / passed
- `cd gradle-plugin && ./gradlew test --tests unit.ApplySpecTest.shouldExposeScala3SourcesOption --tests unit.ApplySpecTest.shouldPassScala3SourcesToScalaGenerators` / passed
- `cd gradle-plugin && ./gradlew test --tests unit.ApplySpecTest.shouldPassScala3SourcesToScalaGenerators` / environment failed once while `sbt javafmtCheckAll` was running concurrently: sbt boot socket Address already in use
- `cd gradle-plugin && ./gradlew test --tests unit.ApplySpecTest.shouldPassScala3SourcesToScalaGenerators` / passed after replacing reflection with VarHandle
- `sbt "++ 3.3.8 runtime/publishLocal" "sbt-plugin/scripted scala3/02-scala3-sourcegen"` / passed
- `sbt docs/paradox` / passed
- `sbt headerCreateAll` / passed
- `sbt +headerCheckAll` / passed
- `sbt checkCodeStyle` / passed
- `scalafmt --mode diff-ref=origin/main` / passed
- `scalafmt --list --mode diff-ref=origin/main` / passed
- JDK 17 `sbt javafmtAll` / passed
- JDK 17 `sbt javafmtCheckAll` / passed
- `cd gradle-plugin && ./gradlew test` / environment failed on Java 25 because Gradle 7.6.4 cannot process class file major version 69
- JDK 17 `cd gradle-plugin && ./gradlew test` / passed with `pekkoGrpcTest.baselineVersion=2.0.0-M2-43-6b2d34c7-20260703-0040-SNAPSHOT`
- `git diff --check` / passed

### References
Fixes #766

### DIFF
--- a/codegen/src/main/scala/org/apache/pekko/grpc/gen/scaladsl/ScalaCompatConstants.scala
+++ b/codegen/src/main/scala/org/apache/pekko/grpc/gen/scaladsl/ScalaCompatConstants.scala
@@ -18,8 +18,11 @@
 package org.apache.pekko.grpc.gen.scaladsl
 
 private[scaladsl] class ScalaCompatConstants(emitScala3Sources: Boolean = false) {
-  // val WildcardType: String = if (emitScala3Sources) "?" else "_"
+  val WildcardType: String = if (emitScala3Sources) "?" else "_"
   val WildcardImport: String = if (emitScala3Sources) "*" else "_"
 
   val ImplicitUsing: String = if (emitScala3Sources) "using " else ""
+  val ImplicitParameter: String = if (emitScala3Sources) "using" else "implicit"
+  val ImplicitVal: String = if (emitScala3Sources) "given" else "implicit val"
+  val ImplicitDef: String = if (emitScala3Sources) "given" else "implicit def"
 }
--- a/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
@@ -53,15 +53,15 @@ trait @{service.name}Client extends @{service.name} with @{service.name}ClientPo
 
 @@PekkoGrpcGenerated
 object @{service.name}Client {
-  def apply(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider): @{service.name}Client =
+  def apply(settings: GrpcClientSettings)(@{service.scalaCompatConstants.ImplicitParameter} sys: ClassicActorSystemProvider): @{service.name}Client =
     new Default@{service.name}Client(GrpcChannel(settings), isChannelOwned = true)
-  def apply(channel: GrpcChannel)(implicit sys: ClassicActorSystemProvider): @{service.name}Client =
+  def apply(channel: GrpcChannel)(@{service.scalaCompatConstants.ImplicitParameter} sys: ClassicActorSystemProvider): @{service.name}Client =
     new Default@{service.name}Client(channel, isChannelOwned = false)
 
-  private class Default@{service.name}Client(channel: GrpcChannel, isChannelOwned: Boolean)(implicit sys: ClassicActorSystemProvider) extends @{service.name}Client {
+  private class Default@{service.name}Client(channel: GrpcChannel, isChannelOwned: Boolean)(@{service.scalaCompatConstants.ImplicitParameter} sys: ClassicActorSystemProvider) extends @{service.name}Client {
     import @{service.name}.MethodDescriptors.@{service.scalaCompatConstants.WildcardImport}
 
-    private implicit val ex: ExecutionContext = sys.classicSystem.dispatcher
+    private @{service.scalaCompatConstants.ImplicitVal} ex: ExecutionContext = sys.classicSystem.dispatcher
     private val settings = channel.settings
     private val options = NettyClientUtils.callOptions(settings)
 
--- a/codegen/src/main/twirl/templates/ScalaCommon/Marshallers.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaCommon/Marshallers.scala.txt
@@ -36,18 +36,18 @@ import pekko.grpc.PekkoGrpcGenerated
 @@PekkoGrpcGenerated
 object @{service.name}Marshallers {
   @for(serializer <- service.serializers) {
-  implicit val @serializer.name: ScalapbProtobufSerializer[@{serializer.scalaType}] = @{service.packageName}.@{service.name}.Serializers.@{serializer.name}
+  @{service.scalaCompatConstants.ImplicitVal} @serializer.name: ScalapbProtobufSerializer[@{serializer.scalaType}] = @{service.packageName}.@{service.name}.Serializers.@{serializer.name}
   }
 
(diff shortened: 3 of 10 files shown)


pjfanning (approved):
lgtm

PR #778, opened 2026-07-04
refactor: remove deprecated scalaPartialFunction (since 1.2.0)

He-Pin:
### Motivation
`scalaPartialFunction` was deprecated since pekko-grpc 1.2.0 as it is no longer needed since support for Scala 2.11 was dropped.

### Modification
Remove the deprecated `scalaPartialFunction` method and inline its logic into `scalaAnonymousPartialFunction`. Remove unused `@nowarn` import.

### Result
Cleaner API surface with one less deprecated method.

### Tests
Not run - trivial inline refactor

### References
None - deprecated API cleanup


pjfanning:
This was only recently deprecated and I prefer to keep deprecated code around for a while.

PR #820, opened 2026-07-30
add verify-hostname support

pjfanning:
Problem

The Pekko HTTP client backend (PekkoHttpClientUtils) never called setEndpointIdentificationAlgorithm("HTTPS") on the SSLEngine. This meant no RFC 2818 hostname verification — the client would accept any valid certificate regardless of hostname, enabling MitM attacks. The Netty backend was not affected (gRPC-shaded Netty verifies by default).

Solution: Opt-in config with warning

reference.conf — new config key:
hocon
```
verify-hostname = false
```
Default is false (preserves existing behavior). The warning in the log nudges users toward enabling it.

GrpcClientSettings — new field and setter:
- verifyHostname: Boolean — read from config
- withVerifyHostname(value: Boolean) — programmatic override

PekkoHttpClientUtils — conditional verification:
- When verifyHostname = true: sets setEndpointIdentificationAlgorithm("HTTPS") on the SSLEngine, enabling hostname checking
- When verifyHostname = false: skips the setting (existing behavior), logs a warning once per channel creation:
▎ TLS hostname verification is disabled for pekko-http client 'X'. This is insecure and should only be used for testing. Enable it with verify-hostname = true in your configuration. Note: the netty backend always verifies hostnames.

Usage

hocon
```
pekko.grpc.client."*" {
  verify-hostname = true
}
```

Or programmatically:
```
val settings = GrpcClientSettings.connectToServiceAt("localhost", 8080)
  .withVerifyHostname(true)
```

### DIFF
--- a/runtime/src/main/resources/reference.conf
+++ b/runtime/src/main/resources/reference.conf
@@ -38,6 +38,12 @@ pekko.grpc.client."*" {
   # leave empty to auto-detect, or configure 'jdk' or 'openssl'.
   ssl-provider = ""
 
+  # Whether to verify the server's hostname against its TLS certificate (RFC 2818).
+  # When false (the default), the client accepts any valid certificate regardless
+  # of hostname. This is insecure for production and should only be used for testing.
+  # Only effective for the pekko-http backend; the netty backend always verifies.
+  verify-hostname = false
+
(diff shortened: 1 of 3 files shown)


He-Pin (approved):

raboof:
> Default is false (preserves existing behavior)

I think the default should be 'true' (especially since the next release will be a major version bump anyway)

pjfanning:
I'll work on a follow up PR that changes the config to default to true

PR #822, opened 2026-07-31
trailer frame support in GrpcProtocol appears broken

pjfanning:
AbstractGrpcProtocol.scala — Negative frame length validation
- Added if (length < 0) check in both the strict decoder (line 137) and streaming decoder (line 157)
- Prevents undefined behavior in ByteReader.take() when the high bit of the 4-byte length field is set

GrpcProtocolWeb.scala — Three gRPC-Web bugs fixed:
1. Frame type mask: 80 (decimal, 0x50) → 0x80 (hex, 128). The old mask 0x50 would match trailer frames (0x80) as data frames (0x00 & 0x50 == 0), silently swallowing trailers.
2. Match arm: case 1 → case 0x80. With the corrected mask, trailer frames yield 0x80, not 1. The old case never matched.
3. decodeTrailer: replaced ??? (runtime NotImplementedError) with actual parsing of key:value\r\n format back into List[HttpHeader].

GrpcProtocolWebSpec.scala — New test file with 4 tests:
- Data frame encode/decode roundtrip
- Trailer frame encode/decode roundtrip
- Data vs trailer frame discrimination by type bit
- Negative frame length rejection

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcProtocolWeb.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcProtocolWeb.scala
@@ -92,23 +92,27 @@ abstract class GrpcProtocolWebBase(subType: String) extends AbstractGrpcProtocol
         (str.charAt(i) == ' ' || str.charAt(i) == '\t' || str.charAt(i) == '\r' || str.charAt(i) == '\n'))
         i += 1
       if (i >= len) return headers.result()
-      // scan for colon (key:value separator)
+      // scan for colon (key:value separator), stopping at LF for malformed lines
       val keyStart = i
-      while (i < len && str.charAt(i) != ':') i += 1
-      if (i >= len) return headers.result() // no colon found, malformed
-      val keyEnd = i
-      i += 1 // skip ':'
-      // scan for LF (line terminator)
-      val valueStart = i
-      while (i < len && str.charAt(i) != '\n') i += 1
-      var valueEnd = i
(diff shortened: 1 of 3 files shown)


runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcProtocolWeb.scala
```diff
@@ -80,7 +81,41 @@ abstract class GrpcProtocolWebBase(subType: String) extends AbstractGrpcProtocol
...
+    val str = data.utf8String
+    val len = str.length
+    val headers = List.newBuilder[HttpHeader]
+    var i = 0
+    while (i < len) {
+      // skip leading whitespace and blank lines
+      while (i < len &&
+        (str.charAt(i) == ' ' || str.charAt(i) == '\t' || str.charAt(i) == '\r' || str.charAt(i) == '\n'))
+        i += 1
+      if (i >= len) return headers.result()
```
Philippus: I'd prefer that we don't use `return`. You can use a `done` boolean to bail out of the loop for example.

Philippus (changes requested):

runtime/src/main/scala/org/apache/pekko/grpc/internal/GrpcProtocolWeb.scala
pjfanning: @Philippus I refactored the code to remove the return

Philippus (approved):

PR #824, opened 2026-07-31
apply deadline on pekko-http backend

pjfanning:
The Pekko HTTP backend now applies gRPC deadlines:

- Already expired — fails immediately with DEADLINE_EXCEEDED and the overshoot duration
- Active deadline — wraps the response source with completionTimeout; if the stream doesn't complete in time, fails with DEADLINE_EXCEEDED
- No deadline set (null) — no change to behavior

The Netty backend already handled deadlines natively through gRPC's Netty channel. This brings the Pekko HTTP backend to parity.

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoHttpClientUtils.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoHttpClientUtils.scala
@@ -37,6 +37,7 @@ import io.grpc.{ CallOptions, MethodDescriptor, Status, StatusRuntimeException }
 import javax.net.ssl.{ KeyManager, SSLContext, TrustManager }
 import scala.collection.immutable
 import scala.concurrent.{ ExecutionContext, Future, Promise }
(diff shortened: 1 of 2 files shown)


He-Pin (approved):

runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoHttpClientUtils.scala
```diff
@@ -185,7 +185,37 @@ object PekkoHttpClientUtils {
             descriptor.getFullMethodName),
           GrpcEntityHelpers.metadataHeaders(headers.entries),
           source)
-        responseToSource(singleRequest(httpRequest), deserializer)
+        val base = responseToSource(singleRequest(httpRequest), deserializer)
+        applyDeadline(base, options)
```
He-Pin: Maybe just inline it, remove the base

pjfanning: done

PR #830, opened 2026-08-05
Verify hostname defaulting to true

pjfanning:
follow up to #820

### DIFF
--- a/runtime/src/main/resources/reference.conf
+++ b/runtime/src/main/resources/reference.conf
@@ -39,10 +39,10 @@ pekko.grpc.client."*" {
   ssl-provider = ""
 
   # Whether to verify the server's hostname against its TLS certificate (RFC 2818).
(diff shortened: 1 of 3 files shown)


runtime/src/main/resources/reference.conf
```diff
@@ -39,10 +39,10 @@ pekko.grpc.client."*" {
   ssl-provider = ""
 
   # Whether to verify the server's hostname against its TLS certificate (RFC 2818).
-  # When false (the default), the client accepts any valid certificate regardless
-  # of hostname. This is insecure for production and should only be used for testing.
+  # When false, the client accepts any valid certificate regardless of hostname.
+  # This is insecure and should only be used for testing.
```
raboof, line 43: (TBH there's scenario's where this is fine, but indeed that should be a conscious operator decision)

raboof (approved):

PR #842, opened 2026-08-18
Override newWriter and newReader in GrpcProtocolNative

Philippus:
cherry-picked from https://github.com/akka/akka-grpc/pull/1787, which is now Apache 2.0.


pjfanning:
This change is probably ok. The other changes that you left out from 1787 turned out to be problematic, causing issues with gzip encoding.
What is the main purpose of this remaining change?

PR #850, opened 2026-08-20
Fix Codecs.negotiate() dropping supported encodings with whitespace

### LINKED ISSUE #847: Codecs.negotiate() fails to parse comma-separated grpc-accept-encoding values, silently falling back to Identity for most gRPC clients
## Summary

`Codecs.negotiate()` in `pekko-grpc-runtime` does not split comma-separated values in a single `grpc-accept-encoding` header occurrence. The result is that virtually every gRPC client other than grpc-java receives uncompressed responses, regardless of whether the server is configured to compress.

- **Version:** 1.1.1
- **Component:** `pekko-grpc-runtime` — `org.apache.pekko.grpc.internal.Codecs`

---

## Root cause

`Codecs.negotiate()` branches on the number of header *occurrences* returned by `Message-Accept-Encoding.findIn(headers)`:

```scala
if (accepted.isEmpty) {
  Identity
} else if (accepted.length == 1) {
  supportedByName.getOrElse(accepted.head, Identity)   // ← bug
} else {
  accepted.collectFirst { case a if supportedByName.contains(a) =>
    supportedByName(a) }.getOrElse(Identity)
}
```

Because pekko-http has no registered grammar for `grpc-accept-encoding`, `findIn` treats it as a raw custom header and returns one element per header *line*. The standard client wire format is a single line with a comma-separated list:

```
grpc-accept-encoding: identity, deflate, gzip
```

This produces `accepted = Seq("identity, deflate, gzip")` — length 1. The single-occurrence branch then does:

```scala
supportedByName.getOrElse("identity, deflate, gzip", Identity)
```

The whole string is not a key in `supportedByName`, so the lookup misses and the server falls back to `Identity`, serving an uncompressed response even though the client advertised gzip support.

The multi-occurrence branch (`else`) is correct — it checks each element individually — but no standard client sends multiple separate `grpc-accept-encoding` lines.

---

## Why grpc-java appears unaffected

grpc-java sends the header as a bare single token with no comma:

```
grpc-accept-encoding: gzip
```

`accepted.head` is then `"gzip"`, which does match `supportedByName`, so grpc-java accidentally triggers correct behaviour. This is not tolerance of the bug — it is simply never exposed by grpc-java's wire format.

**Affected clients** (confirmed to send the comma-separated form): grpc-python, grpc-go, grpc-node, grpcurl. Likely all non-Java gRPC implementations.

---

## Reproduction

Start any pekko-grpc server with gzip compression enabled and call it from grpc-python:

```python
import grpc
channel = grpc.insecure_channel("localhost:8080")
# grpc-python sends: grpc-accept-encoding: identity, deflate, gzip
```

Observe that the server response carries `grpc-encoding: identity` (or no `grpc-encoding` header) and an uncompressed payload, despite the client advertising gzip. Call the same endpoint from grpc-java and observe that the response is correctly compressed (`grpc-encoding: gzip`, valid 5-byte frame prefix, real gzip payload).

---

## Suggested fix

In the single-occurrence branch, split on commas before the lookup. The two branches can then be unified:

```scala
def negotiate(request: jm.HttpRequest): Codec = {
  val headers = extractHeaders(request)
  val accepted = `Message-Accept-Encoding`.findIn(headers)
  accepted
    .flatMap(_.split(','))
    .map(_.trim)
    .collectFirst { case name if supportedByName.contains(name) => supportedByName(name) }
    .getOrElse(Identity)
}
```

This correctly handles a comma-separated single occurrence, multiple separate occurrences, and the empty case, and it matches the behaviour of the existing multi-occurrence branch for the case that already worked.

---

## Additional context

A related but distinct question was raised in #286 (whether gzip should appear in `Message-Accept-Encoding` by default). This issue is different: it is about the server failing to recognise gzip even when the client explicitly advertises it.

haroldb1:
Motivation:
grpc-accept-encoding headers formatted with a space after the comma (e.g. "deflate, gzip", as sent by grpc-go/grpc-python/grpcurl) were never matched against supported codec names, silently downgrading negotiation to Identity even when the client offered gzip.

Modification:
Trim each split token in `Message-Accept-Encoding`.findIn before codec-name lookup.

Result:
Codecs.negotiate now correctly picks a supported codec regardless of whitespace around commas in grpc-accept-encoding.

Tests:
- sbt "runtime / testOnly org.apache.pekko.grpc.CodecsSpec" - pass

References:
Fixes #847

### DIFF
--- a/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/scaladsl/headers/headers.scala
@@ -44,7 +44,7 @@ object `Message-Accept-Encoding` extends ModeledCustomHeaderCompanion[`Message-A
     Try(new `Message-Accept-Encoding`(value))
 
   def findIn(headers: Iterable[jm.HttpHeader]): Array[String] =
-    headers.collectFirst { case h if h.is(name) => h.value().split(',') }.getOrElse(Array.empty)
+    headers.collectFirst { case h if h.is(name) => h.value().split(',').map(_.trim) }.getOrElse(Array.empty)
 
   /** Java API */
   def findIn(headers: java.lang.Iterable[jm.HttpHeader]): Array[String] = {
--- a/runtime/src/test/scala/org/apache/pekko/grpc/CodecsSpec.scala
+++ b/runtime/src/test/scala/org/apache/pekko/grpc/CodecsSpec.scala
@@ -72,6 +72,11 @@ class CodecsSpec extends AnyWordSpec with Matchers with TryValues {
       Codecs.negotiate(request) should be(Gzip)
     }
 
+    "negotiate gzip when grpc-accept-encoding uses comma+space separators (as sent by grpc-go/grpc-python/grpcurl)" in {
+      val request = HttpRequest(headers = immutable.Seq(RawHeader("grpc-accept-encoding", "deflate, gzip")))
+      Codecs.negotiate(request) should be(Gzip)
+    }
+
   }
 
   "Detecting message encoding from remote" should {


pjfanning (approved):
lgtm - thanks

PR #856, opened 2026-08-22
Update pekkoVersion to 2.0.0-M4

pjfanning:
Had to modify one file because it was using a deprecated method for queue.


runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoHttpClientUtils.scala
```diff
@@ -133,7 +133,7 @@ object PekkoHttpClientUtils {
 
     val (queue, doneFuture) =
       Source
-        .queue[HttpRequest](4242, OverflowStrategy.fail)
+        .queue[HttpRequest](4242)
```
pjfanning, line 136: @He-Pin does this look a good change to the queue?

He-Pin: yes,we should use the bounded queue now.

He-Pin (approved):

PR #845, opened 2026-08-19
reverted: add verify-hostname support (#820)

pjfanning:
cherry pick af8aa19e73ca07995b6b06dce2c85895dff6f726 #820

reverted by #858

PR #858, opened 2026-08-23
Revert verify-hostname support on 1.3.x

pjfanning:
### Motivation

The backport of #820 (#845) makes 1.3.x **less secure than 1.2.x** for the pekko-http client backend. This should be reverted before 1.3.0 ships.

Another AI suggested that verify-hostname was missing but it is implemented by pekko-http and netty for us.

`ConnectionContext.httpsClient(SSLContext)` — which 1.2.x calls — already sets both client mode and the `https` endpoint identification algorithm in **pekko-http 1.1.0**, the version this branch depends on:

```scala
def httpsClient(context: SSLContext): HttpsConnectionContext =
  httpsClient((host, port) => {
    val engine = context.createSSLEngine(host, port)
    engine.setUseClientMode(true)
    engine.setSSLParameters {
      val params = engine.getSSLParameters
      params.setEndpointIdentificationAlgorithm("https")
      params
    }
    engine
  })
```

Hostname verification was therefore never absent on this path. #845 moved to the `httpsClient((host, port) => ...)` overload — which pekko-http marks `@ApiMayChange` and documents as *"it is up to you to make sure features like SNI and hostname verification are enabled as needed"* — and set the endpoint identification algorithm only `if (settings.verifyHostname)`. `reference.conf` on this branch has `verify-hostname = false`, and #830, which flipped that default to `true` on `main`, was never backported.

Net effect on 1.3.x today: a pekko-http backend client with `use-tls = true` accepts **any** certificate its trust store chains to, regardless of which host it was issued for. Upgrading 1.2.x → 1.3.0 would silently lose hostname verification with no configuration change.

Blast radius is limited to the `pekko-http` backend (`netty` is still the default and always verifies) and to TLS connections, and each channel does log a warning — but it is on by default, and a warning is not consent.

### Modification

Restores the three source files touched by #845 to their pre-backport state:

- `runtime/src/main/resources/reference.conf`
- `runtime/src/main/scala/org/apache/pekko/grpc/GrpcClientSettings.scala`
- `runtime/src/main/scala/org/apache/pekko/grpc/internal/PekkoHttpClientUtils.scala`

This is deliberately **not** a revert of the whole commit: #845 also carried unrelated GitHub Actions changes ("upgrade setup-sbt due to build issues") across 11 workflow files, and those are kept untouched.

Two things that make the revert clean:

- `verifyHostname` / `withVerifyHostname` have never appeared in a release, so removing them moves the public API back *towards* the MiMa baseline rather than away from it.
- #845 added no MiMa excludes, so none are left dangling.

The import reordering that #845 incidentally applied to `PekkoHttpClientUtils` is reverted along with it, leaving the file matching 1.2.x.

### Result

1.3.x verifies the server hostname against its certificate again, matching 1.2.x and the pre-backport behaviour. No public API change relative to the 1.2.x line.

For `main`, the equivalent hardening plus regression tests and documentation is in #857; the tests there do not port to this branch as-is, since they exercise `withVerifyHostname` and helpers that only exist once this feature is present.

### Tests

- `sbt "runtime/test"` — 110 tests passed
- `sbt scalafmtCheckAll scalafmtSbtCheck` — passed
- `sbt "sbt-plugin / scripted"` — Not run, left to CI
- `sbt +mimaReportBinaryIssues` — Not run; single-version run passed, cross-version left to CI

### References

Refs #820, Refs #845, Refs #830

### DIFF
--- a/runtime/src/main/resources/reference.conf
+++ b/runtime/src/main/resources/reference.conf
@@ -38,12 +38,6 @@ pekko.grpc.client."*" {
   # leave empty to auto-detect, or configure 'jdk' or 'openssl'.
   ssl-provider = ""
 
-  # Whether to verify the server's hostname against its TLS certificate (RFC 2818).
-  # When false (the default), the client accepts any valid certificate regardless
-  # of hostname. This is insecure for production and should only be used for testing.
-  # Only effective for the pekko-http backend; the netty backend always verifies.
-  verify-hostname = false
-
   # TODO: Enforce HTTP/2 TLS restrictions: https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-9.2
 
   # The number of times to try connecting before giving up.
--- a/runtime/src/main/scala/org/apache/pekko/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/org/apache/pekko/grpc/GrpcClientSettings.scala
@@ -158,8 +158,7 @@ object GrpcClientSettings {
       getOptionalString(clientConfiguration, "user-agent"),
       clientConfiguration.getBoolean("use-tls"),
       getOptionalString(clientConfiguration, "load-balancing-policy"),
-      clientConfiguration.getString("backend"),
-      verifyHostname = clientConfiguration.getBoolean("verify-hostname"))
+      clientConfiguration.getString("backend"))
 
   private def getOptionalString(config: Config, path: String): Option[String] =
     config.getString(path) match {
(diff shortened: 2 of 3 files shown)


samueleresca (approved):
Checked the following:

- Changes to the `workflows/*.yml` are still in place.
- The `ConnectionContext.httpsClient(hostname, port)` is now being replaced with the old `ConnectionContext.httpsClient`
- Verified that https://github.com/apache/pekko-grpc/pull/857 test the expectations and fails when the `connectionContext` just builds  `SSLEngine`  without endpoint identification

PR #859, opened 2026-08-23
Test that the pekko-http client verifies the server hostname (1.3.x)

pjfanning:
### Motivation

Stacked on #858 — this branch contains that revert as its first commit, so **merge #858 first**; this PR will then show only its own commit.

Nothing on this branch covered hostname verification, which is how the regression in #845 reached it unnoticed. The behaviour is not written in this project at all — it comes from `ConnectionContext.httpsClient(SSLContext)` setting the `https` endpoint identification algorithm — so it is exactly the kind of implicit contract a future change can drop silently.

### Modification

**A small extraction, which turned out to be necessary.** `sslContextFor` and `connectionContext` are lifted out of `createChannel`. Pure refactor, no behaviour change — but driving the check through `createChannel` does not work.

I tried the zero-production-change route first. `managedPersistentHttp2` wraps the connection in `PersistentConnection`, which *retries* connection failures and breaks the stream rather than propagating the cause. Probing the layers directly:

```
HTTP/1  localhost   = 200 OK
HTTP/1  127.0.0.1   = SSLHandshakeException: No subject alternative names matching IP address 127.0.0.1 found
HTTP/2  localhost   = (connects, fixture server is HTTP/1 only)
HTTP/2  127.0.0.1   = SSLHandshakeException: No subject alternative names matching IP address 127.0.0.1 found
LIKE-GRPC localhost = NoSuchElementException: head of empty stream
LIKE-GRPC 127.0.0.1 = NoSuchElementException: head of empty stream
```

The last two use the client's exact builder configuration. Both hosts look identical from the caller's side, so a test at that level cannot tell a rejected hostname from any other connection failure. Verification itself works fine on both HTTP/1 and HTTP/2 — it is only the observability through `PersistentConnection` that is lost.

**The test.** `HostnameVerificationSpec` runs real TLS handshakes through the connection context the client builds, against a local server whose certificate carries a single `DNS:localhost` SAN:

| host | expected |
|---|---|
| `localhost` | handshake succeeds, 200 |
| `127.0.0.1` | `SSLHandshakeException` |

Same server, same trust store — so the pair isolates the hostname check from a general trust or TLS problem rather than just asserting "something failed".

**Negative control.** With `connectionContext` temporarily changed to build the engine without endpoint identification, the rejecting case fails and the matching case still passes — confirming the guard bites rather than passing vacuously.

The certificates under `runtime/src/test/resources/certs` are copied from `plugin-tester-scala/src/main/resources/certs` on the `main` branch.

### Result

The verification contract is pinned by a test that fails if the connection context is swapped for one that leaves hostname checking off — which is precisely what #845 did.

### Tests

- `sbt "runtime/test"` — 112 tests passed, 2 of them new
- `sbt scalafmtCheckAll scalafmtSbtCheck` — passed
- negative control as described above
- `sbt "sbt-plugin / scripted"` — Not run, left to CI
- `sbt +mimaReportBinaryIssues` — Not run; single-version run passed, cross-version left to CI

### References

Refs #820, Refs #845, Refs #858


samueleresca (approved):
Approving. See: https://github.com/apache/pekko-grpc/pull/858#pullrequestreview-5002554544

PR #867, opened 2026-08-29
Remove note about server reflection

pjfanning:
The doc is well out of date. On main branch we use v1 of the proto and it is almost identical to the v1alpha.
It's also good to remove the link to an old akka issue.


pjfanning:
This broken link causes our link validation job to fail.

nvollmar (approved):

PR #871, opened 2026-08-30
Add a config to stop grpc-web CORS allowing credentials from any origin

pjfanning:
### Motivation

`WebHandler.defaultCorsSettings` sets `withAllowCredentials(true)` and leaves
`allowed-origins` at the pekko-http-cors reference default of `"*"`. A POST from
an arbitrary origin is answered with:

```
Access-Control-Allow-Origin: https://evil.example
Access-Control-Allow-Credentials: true
```

so a browser sends cookies and HTTP authentication on a cross-origin call and
lets the calling page read the response. For a service authenticated that way,
any page a user visits can act as them — a CSRF vector.

Verified against `main` before changing anything; the output above is what the
current default actually returns. Note there is no backend to switch to that
avoids this: grpc-netty in this repo is client-only, so every gRPC-Web
deployment is served by pekko-http through these settings.

### Modification

Added `pekko.grpc.server.grpc-web.allow-credentials-from-any-origin`, default
`true`, so existing deployments are unchanged. While it is `true` and the
combination is actually in effect, each handler logs a warning naming the two
ways out.

Set it to `false` and the credentials permission is dropped — the handler
answers `Access-Control-Allow-Origin: *` with no `Access-Control-Allow-Credentials`,
which is what makes browsers withhold cookies on cross-origin calls.

The check tests `HttpOriginMatcher.*` by identity, which is what
`allowed-origins = "*"` parses to. Settings that list origins are a deliberate
choice, so they are neither warned about nor modified — credentials keep working
once the caller has said who may use them.

Both entry points are covered. The javadsl handler builds its own CORS route and
would otherwise have bypassed this; `scaladsl.CorsSettings` extends the javadsl
one, so it reaches the same policy through a match.

### Result

The insecure combination is reported wherever it is live, and can be switched off
with configuration rather than a code change. The default is unchanged behaviour.

### Tests

`WebHandlerCorsSpec` asserts the response headers in both modes, for actual and
preflight requests:

| setting | `Allow-Origin` | `Allow-Credentials` |
|---|---|---|
| `true` (default) | echoes the caller's origin | `true` |
| `false` | `*` | absent |
| `false`, origins restricted | the restricted origin | `true` |

Confirmed the guard bites: not applying the policy fails 2 tests.

- `sbt "runtime/testOnly ...WebHandlerCorsSpec"` — 8 passed
- `sbt docs/paradox` — passed
- `sbt scalafmtAll scalafmtSbt` — applied
- `sbt "runtime/test"` (full suite) — not run locally, left to CI

### Notes

The default is deliberately left at the insecure-but-compatible value, per
maintainer direction: this ships as an opt-in with a warning rather than a
breaking change. Flipping the default to `false` would be the stronger fix and
is worth considering for a major version, since `WebHandler` is `@ApiMayChange`
and gRPC-Web is documented as experimental.

### DIFF
--- a/docs/src/main/paradox/server/grpc-web.md
+++ b/docs/src/main/paradox/server/grpc-web.md
@@ -49,3 +49,50 @@ Scala
 
 Java
 :  @@snip [Main.java](/plugin-tester-java/src/main/java/example/myapp/CombinedServer.java) { #grpc-web }
+
+## CORS and credentials
+
+@@@ warning
+
+`WebHandler.defaultCorsSettings` allows **any** origin and sets
+`Access-Control-Allow-Credentials: true`. A browser will therefore send cookies and HTTP
+authentication on a cross-origin call to your service, and let the calling page read the
+response. For a service authenticated that way, any web page your users visit can act as
+them — a CSRF vector.
+
+@@@
+
+Restrict the origins that may use credentials:
+
+Scala
+:   ```scala
+    import org.apache.pekko.http.cors.scaladsl.model.HttpOriginMatcher
+    import org.apache.pekko.http.scaladsl.model.headers.HttpOrigin
+
+    implicit val corsSettings: CorsSettings =
+      WebHandler.defaultCorsSettings
+        .withAllowedOrigins(HttpOriginMatcher(HttpOrigin("https://example.com")))
+    ```
+
+Java
+:   ```java
+    CorsSettings corsSettings = WebHandler.defaultCorsSettings()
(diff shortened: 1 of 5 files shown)


timw:
I'd be in favour of just making a breaking change on this in 2.0 TBH.

From a secure by default perspective, `HttpOriginMatcher.* and `withAllowCredentials(true)` are both bad defaults TBH - requiring users to explicitly opt-in to both would be good.

i.e. hosting a public and/or authenticated gRPC service should both be explicit opt-ins for the application, rather than relying (probably unknowingly) on the platform defaults.

